PR #341, opened 2015-01-05
[CDF-334] - Compilation / Minification

afrjorge:
- jQuery needs to be loaded before Bootstrap, fix load issue by adding bootstrap libs to the (compression) exclude list
	- Avoid compressing Sprintf, Utils and Utf8Encoder because they need to be accessible to other AMD modules (e.g. AddIns) that don't depend on a main dashboard module (e.g. Dashboard.Blueprint)


afrjorge:

PR #342, opened 2015-01-27
[CDF-468] [BA-Server EE 4.8.3] Analyzer plugin breaks using CDF core-ref...

pedrofvteixeira:
...actored

	- cdf-pentaho ( 4.8.x ) was including js-lib/ccc/ccc instead of proper js-lib/ccc/pen
	- including js-lib/ccc/pen also requires js/lib/pen-shim.js dependency inclusion
	- pen-shim.js ( in minified cdf-blueprint ) must be declared *before* cdf-core core-scripts-2's def.js, otherwise it will fail with 'pen is not defined'
	- cdf-pentaho ( 4.8.x ) build.xml now includes 'extra-core-scripts 1/2/3', to allow cdf-pentaho to add its own dependencies at cdf-core's core-scripts level
	- updated ccc.js to properly work with new js-lib/ccc/pen ( which in all reality ended up editing build.properties and make it similar to cdf-pentaho5 one )
	- tested core-components , samples and analyzer

### DIFF
--- a/cdf-pentaho/build.properties
+++ b/cdf-pentaho/build.properties
@@ -15,10 +15,10 @@ dependency.jxl.revision=2.6.12
 dependency.junit.revision=4.10
 
 #modifications needed for core-refator components/ccc.js
-cdf.ccc.require.header=
-cdf.ccc.published.globally=
-cdf.ccc.var.declaration=var
-cdf.ccc.require.footer=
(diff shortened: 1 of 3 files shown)


pedrofvteixeira:
@pamval , @dcleao please review

PR #347, opened 2015-02-04
[CDE-447] - ViewManager custom component fails to save/load views

afrjorge:
- Reviewed ViewsApi RESTful API for version 5.x
	- Allow Dashboards.Main.js restoreView() to assign an empty object to view.params
	- Merged ViewEngine and ViewsEngine to avoid code duplication
	- Reviewed unit tests
	- JsonUtil now uses enum instead of public static final strings


afrjorge:
@pamval please review

PR #351, opened 2015-02-06
[CDF-472] - Optimize TableComponent performance

jvelasques:


jvelasques:
@pamval please review

PR #353, opened 2015-02-06

afrjorge:
- Fixed references to deprecated Dashboards object in RefreshEngine
	- Updated Dashboard.views according to the new RESTful Views API
	- Included Base64 in the require lib cfg file
	- updated SESSION_LOCALE in context.js used during unit tests


afrjorge:

PR #354, opened 2015-02-11
[CDF-472] - Optimize TableComponent performance, adding change in TableC...

jvelasques:
...omponent.js


jvelasques:
@pamval please review

PR #361, opened 2015-02-19

afrjorge:
- Remove cache-buster when running in debug mode


afrjorge:

PR #374, opened 2015-02-26

afrjorge:
- Added missing return


afrjorge:
@pamval please review

PR #382, opened 2015-03-11
[CDF-331] - Removing jquery.sparkline as a TableComponent dependency

jvelasques:


jvelasques:

PR #386, opened 2015-03-12
[CDE-471] On Popup Component Reference dragging first popup results in...

davidmsantos90:
... it jumping to bottom of page

This is the ticked where the bug, in jquery.ui that is causing this issue, is discussed http://bugs.jqueryui.com/ticket/9315 

@pamval please review


pamval:
Same for the require-js branch, please.

PR #389, opened 2015-03-13
[CDF-331] - Fixing a bug in ButtonComponent where the scope of label (wh...

jvelasques:
...en a function) was not the expected


jvelasques:

PR #397, opened 2015-03-17
[CDF-334] - Compilation / Minification

afrjorge:
- simplify run-karma ant target by moving require file compilation into test-js-require ant target


afrjorge:
@pamval please review

PR #400, opened 2015-03-19

afrjorge:
cherry-picked latest commits from buildguy [ENGOPS-999]


afrjorge:

PR #405, opened 2015-03-20
[CDF-465] - Normalizing locale before initiating i18n

jvelasques:


jvelasques:

PR #407, opened 2015-03-23
[CDF-495] - Fix for javascript inside href (Firefox)

jvelasques:


jvelasques:

PR #413, opened 2015-03-26
[CDF-519] - Change style loading mechanism in CDF

jvelasques:


jvelasques:
@pamval @afrjorge @diogofscmariano please review

PR #418, opened 2015-03-30
[CDF-330] - Recovering some tests + fixing minor issues with some module...

jvelasques:
...s


jvelasques:
@pamval please review

PR #420, opened 2015-03-30
[SP-1771] - Backport of PRD-5398 - Prompt screen flashes and take user b...

MikhailHubanau:
...ack to the top again (5.3 Suite)

store scroll position to cookie and restore after refresh


MikhailHubanau:
@pamval @davidmsantos90 would you mind to review and merge?

pamval:
@MikhailHubanau, I'm not sure this is the right place for this fix. This issue affects the prompting library only and not every CDF dashboard. 

Isn't there a  a change to prompting that we could try to fix this ?

MikhailHubanau:
Fix for original issue https://github.com/webdetails/cdf/pull/421

MikhailHubanau:
@pamval, I tested this fix with report attached in http://jira.pentaho.com/browse/PRD-5398.

I was trying to solve this issue on pentaho-platform-plugin-reporting side, but I had to move the solution on CDF layer, because CDF logic (in our case) wraps plugin-reporting code and discards efforts spent on plugin-reporting side. 

Does it make sense?

MikhailHubanau:
Searching better solution.

PR #422, opened 2015-04-01
[CDF-507] Some code is breaking in Internet Explorer 8 because it does ...

davidmsantos90:
...not natively support String.prototype.trim()


pamval:
@davidmsantos90 rebase and make sure this fix applies to legacy and require-js

davidmsantos90:
@pamval rebased and working on both require-js and legacy. please review

PR #428, opened 2015-04-08
[CDF-328][CDF-330]

afrjorge:
- Avoid running JS unit tests only in localhost port 9876, allow other ports to be used
	- Run base JS unit tests with core ant target test-js
	- Avoid using cache buster for unit tests
	- Simplify some regex in build.xml


afrjorge:

PR #431, opened 2015-04-09
[CDF-525] - Charts don't render in require dashboards in IE8

jvelasques:


jvelasques:
@pamval please review

cdf-core/cdf/js-modules/components/ProtovisComponent.js
```diff
@@ -13,9 +13,10 @@
 
 define([
     './ChartComponent',
+    '../lib/jquery',
     '../lib/CCC/protovis',
-    '../lib/jquery'],
-  function(ChartComponent, pv, $) {
+    '../lib/CCC/protovis-msie'],
```
dcleao: It looks like we never noticed it because we always test in non-IE or non-debug. Analyzer and other Pentaho plugins include the msie file in the compressed files, so we never realized that the AMD dependencies of the pvc module do not contain protovis-msie. Ideally, pvc would declare a "conditional dependency" by using some smart browser-version-test AMD plugin.
Anyway, should we change this directly at the source?

PR #433, opened 2015-04-10
[CDF-328][CDF-492][CDF-329] - Move external libs to dedicated folder

afrjorge:
- legacy JS must avoid using libs that register themselves as AMD modules, adding pre-require versions to legacy source-code


afrjorge:
@pamval @diogofscmariano @jvelasques please review, somehow this got lost during the merge to master

PR #439, opened 2015-04-10
[CLEANUP]

afrjorge:
- Avoid DISCONNECTED messages while running tests using PhantomJS, for 4.x and 5 tests also


afrjorge:
@pamval @diogofscmariano @jvelasques please review, also applied changes to 4.x and 5 tests

PR #440, opened 2015-04-13
[CDF-495] - Fixing a sample, javascript inside href in Firefox needs to ...

jvelasques:
...return undefined


jvelasques:

PR #449, opened 2015-04-17
[CDF-531] AutocompleteBoxComponent doesn't refresh the list when removing...

davidmsantos90:
... a character

@pamval please review


pamval:
Why are we changing a 3rd party library ? Are we upgrading version or something ?

davidmsantos90:
While thinking of a solution for this issue, I found this lines in that library:
 - https://github.com/webdetails/cdf/blob/master/cdf-core/cdf/js-lib/autobox/jquery.ui.autobox.js#L93
 - https://github.com/webdetails/cdf/blob/master/cdf-core/cdf/js-lib/autobox/jquery.ui.autobox.js#L264
 - https://github.com/webdetails/cdf/blob/master/cdf-core/cdf/js-lib/autobox/jquery.ui.autobox.js#L528

Indicating that changes were made previously in this file.

@pamval should I try a different solution?

PR #456, opened 2015-04-24
[SP-1771] Backport of PRD-5398 - Prompt screen flashes and take user back to the top again

MikhailHubanau:
Should be merged along with https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/365

@dcleao @rfellows 
Please review.


dcleao:
I think we should first try to solve the problem by changing Common-UI only.

PR #460, opened 2015-04-30
[CDF-531] updated autoComplete component to use jquery.ui autocomplete

davidmsantos90:
@pamval please review


pamval:
if we're no longer using jquery.ui.autobox, we should remove the library

davidmsantos90:
@pamval removed library and all references to it in CDF. please review

pamval:
Merge only after 5.4 branching

PR #463, opened 2015-04-30
BACKLOG-3056 -- Checkin Localization files from Localization Company

kolinus:
What was done:
1) added files provided by the company.


kolinus:
@pamval  could you please review

cdf-core/resource/resources/languages/messages_supported_languages.properties
```diff
@@ -1,3 +1,6 @@
 en=English
 hr=Croatian
 it==Italian
+fr=Francais
+de=Deutsch
+ja=Japanese
```
pamval: This should read ja=日本語. This is for all plugins/projects.

PR #466, opened 2015-05-06
[CDF-491] Remove GetHeaders endpoint from CDF's public api

davidmsantos90:
@pamval please review


pamval:
Merge only after 5.4 branching

PR #470, opened 2015-05-08
[CDE-521] - Mapping jquery usages inside cdf to the correct module

jvelasques:


jvelasques:
@pamval please review

PR #473, opened 2015-05-13
[CDF-494] Incorporate Template Component into CDF

davidmsantos90:
- upgraded mustache in legacy from 0.5.0-dev to 0.5.2 (needed for mustache wax)
 - started using mustache-wax lib v0.9.0 (mustache formatters)


pamval:
Merge only after 5.4 branching.

Need two cases in the New Component process - one for mustache, the other for mustache-wax

pamval:
need to rebase this one.

PR #475, opened 2015-05-13
Eric Hynds Multiselect lib updated

marpontes:
There was an issue with the latest versions of CDF and this JQuery
Multiselect plugin [Hynds].

If you had a multiselect rendered with this plugin, it would popup to
show the options on the visualization, but it would never hide again.

So the core js file was updated to its latest version and the dropdown
menu works like a charm.

### DIFF
--- a/cdf-core/cdf/js-lib/hynds/jquery.multiselect.js
+++ b/cdf-core/cdf/js-lib/hynds/jquery.multiselect.js
@@ -1,670 +1,735 @@
-/*!
-* jQuery MultiSelect UI Widget 1.12
-* Copyright (c) 2011 Eric Hynds
-*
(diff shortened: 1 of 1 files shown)


pamval:
To merge after 5.4 branching.

Needs a new component issue in jira due to external library update.

shivasakthi18:
Any resolution for this. ?

pamval:
The external library update has not been accepted on our internal external component process. We won't be processing this as is. We'll probably develop a way to allow external developers to plug different types of selectors - That will be acceptable, but we're not able to bundle this new version in product.

pamval:
see http://jira.pentaho.com/browse/CDF-790 for reference

PR #483, opened 2015-05-18
PDB-1830 - Regression: Dashboard Text Field prompt ignoring Width spe…

Aliaksandr-Kastenka:
…cified


Aliaksandr-Kastenka:
@pamval 
Please review.

pamval:
Aliaksandr, would be safer if we support both properties for setting width and maxChars.

Something like 

(myself.size ? ("' size='" + myself.size) : (myself.charWidth ?  ("' size='" + myself.charWidth : "" ) ).

This will avoid regressions. 
Ideally, we would output a log warning the user that size and maxLength are deprecated, if they are still using it.

Aliaksandr-Kastenka:
Will update, please do not merge it. We got code freeze.

Aliaksandr-Kastenka:
@pamval 
Please review.

PR #485, opened 2015-05-22
[CDF-494] Incorporate Template Component into CDF

davidmsantos90:
@pamval please review


pamval:
mustache 0.8.2 is already used by the require-js version of CDF. Can't we use that instead of 0.5.2 ?
@afrjorge, why did we downgrade mustache version for legacy js ?

davidmsantos90:
Removed mustache 0.5.2 from legacy and started using mustache 0.8.2 in both versions
@pamval please review

PR #486, opened 2015-05-25
[CDF-558]

diogofscmariano:
- Updated CDF endpoints to allow cross domain requests used in embedded scenarios. Added property to allow configure this use case. By default it doest not insert http headers for cross domain calls
- Fixed getJSON calls that were not defining the xhr fields to allow cross domain requests.
- Added unit tests for the new CorsUtil

@pamval @jvelasques @afrjorge please review

### DIFF
--- a/cdf-core/cdf/js-lib/cdf-core-lib-require-js-cfg.js
+++ b/cdf-core/cdf/js-lib/cdf-core-lib-require-js-cfg.js
@@ -47,15 +47,24 @@
 
     if(!isDebug) { requireCfg.urlArgs = "ts=" + (new Date()).getTime(); }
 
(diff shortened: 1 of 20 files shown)


diogofscmariano:
@pamval @jvelasques @afrjorge please review

cdf-pentaho5/resource/settings.xml
```diff
@@ -23,6 +23,13 @@
 
         accepted values are: true | false
      -->
-    <legacy-dashboard-context>false</legacy-dashboard-context>    
+    <legacy-dashboard-context>false</legacy-dashboard-context>
+
+    <!--
```
pamval: Document this flag - do not push changes to settings.xml.

Make sure code works without the flag in settings.xml and that it defaults to false.

diogofscmariano:
@pamval @jvelasques @afrjorge please review

PR #496, opened 2015-05-28
[CDF-569] - CDF should allow resources to be served by cache on reload

afrjorge:
- disabled cache buster for AMD modules


afrjorge:
@pamval please review

PR #498, opened 2015-05-29
[CDF-548] When updating a table component, with an open expanded row,…

davidmsantos90:
… further expands on the updated component are always empty

@pamval please review


davidmsantos90:
Now a clone of the object to expand is kept inside the Table Component, removing the need to reattach it to the DOM after a row collapse.

@pamval please review

pamval:
does cloning work with parameters and listeners ? 
I mean, if the element you're setting on the expanded object has cdf components, wired with listeners and parameters, does that work properly with the clone ?

davidmsantos90:
In .clone() documentation it says that if you pass the flag true in the argument it will keep all data and events associated to the original.

I tested with a table component being expanded inside the main table with listeners and parameters and all was being triggered correctly.

PR #502, opened 2015-06-01
[CDE-551] - Creating a table with Sparkline column, it's tooltips are…

jvelasques:
… not well formatted


jvelasques:
@pamval please review

PR #504, opened 2015-06-04
[CLEANUP]

afrjorge:
- deleting ${js.build.ouput.dir} in build.xml is irrelevant, this directory is kept/recreated according to the property keepBuildDir in the optimisation config file (default is false)


afrjorge:
@pamval please review

PR #506, opened 2015-06-09
[CLEANUP][CDF-494]

afrjorge:
- removed unused dependency underscore from Dashboard.storage.js
	- replaced _.extend with $.extend (both create shallow copies and jQuery does it faster) in Dashboard.i18n.js, removed underscore dependency
	- check style, unterminated statements, unnecessary semicolons


afrjorge:
@pamval please review

afrjorge:

PR #513, opened 2015-06-17
Merge remote-tracking branch 'upstream/future-develop'

afrjorge:
@pamval please review, not sure if the history will be incorrectly changed. @dcleao will be reviewing the failing unit tests that are temporarily disabled.


pamval:
this is rewriting history and bringing back a lot of commits we don't want to keep.

Like I said earlier today, I think it's safer to cherry-pick future-develop changes (those should be fairly constrained and small).

PR #521, opened 2015-06-24
[CDF-236] Clicking many times on a Button Component in a short period…

plagoa:
… of time triggers many server-side actions

updated: added new verification for enabling the button when there's no expression present, updated the tests

### DIFF
--- a/cdf-core/cdf/js-modules/components/ButtonComponent.js
+++ b/cdf-core/cdf/js-modules/components/ButtonComponent.js
@@ -22,6 +22,8 @@ define(['amd!../lib/underscore', '../lib/jquery', './ActionComponent'],
     render: function() {
       var myself = this;
       var b = $("<button type='button'/>")
+        .addClass('buttonComponent')
+        .addClass('enabled')
         .text(typeof myself.label === 'function' ? myself.label() : myself.label)
         .unbind("click")
         .bind("click", function() {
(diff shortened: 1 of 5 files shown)


plagoa:
@pamval please review

cdf-core/cdf/js-modules/components/ButtonComponent.js
```diff
@@ -45,6 +75,7 @@ define(['amd!../lib/underscore', '../lib/jquery', './ActionComponent'],
      */
     disable: function() {
       this.placeholder('button').attr('disabled', 'disabled');
+      this.placeholder('button').css({ opacity: 0.5 });
     },
```
pamval: Probably best if we assign a css class to the button and have this style defined in cdf.css (not sure about the name). That way it can be overriden by implementors

cdf-core/cdf/js-modules/components/ButtonComponent.js
```diff
@@ -52,6 +83,7 @@ define(['amd!../lib/underscore', '../lib/jquery', './ActionComponent'],
      */
     enable: function() {
       this.placeholder('button').removeAttr('disabled');
+      this.placeholder('button').css({ opacity: 1 });
     },
```
pamval: Probably best if we assign a css class to the button and have this style defined in cdf.css (not sure about the name). That way it can be overriden by implementors

plagoa:
@pamval please review

PR #525, opened 2015-07-02
[CDF-579] Integrate new Selector component into CDF

pedrofvteixeira:
- fixed 'alwaysExpanded' behaviour
		- was: 'alwaysExpanded' became the active behaviour *after* the first onClick event in the component
		- now: 'alwaysExpanded' behaviour  active right at initialize/render


pedrofvteixeira:
@pamval , @carlosrusso please review

carlosrusso:
Seems fine to me.

PR #526, opened 2015-07-02
[CDF-236] Clicking many times on a Button Component in a short period of time triggers many server-side actions

afrjorge:
- moved successCallback and errorCallback outside of the click callback function, avoids adding a new closure function per click
	- added missing CSS styling in CDF 5.x legacy version
	- fixed ReferenceError in unit test


afrjorge:
@pamval please review

PR #527, opened 2015-07-02
[CDF-236] added validation for undefined successCallback and / or fai…

plagoa:
…lureCallback


plagoa:
@pamval please review

PR #534, opened 2015-07-08
[CDF-591] - When updating a table component, with an open expanded ro…

jvelasques:
…w, the current expand is always empty


jvelasques:
@pamval please review

PR #541, opened 2015-07-13

afrjorge:
- added void(0) in the end of <url-template> CDATA section so firefox doesn't open a new window/tab when bar chart is clicked


afrjorge:
@pamval please review

pamval:
cherry-pick to 5.x please

PR #542, opened 2015-07-13

afrjorge:
- applying the same fix to the sample "pentaho_sample" referred in the end of the issue description section


afrjorge:
@pamval please review

PR #545, opened 2015-07-13

afrjorge:


afrjorge:
@pamval please review, cherry-picks

PR #547, opened 2015-07-17
[CDF-598] FilterComponent deletes its placeholder when it is updated …

pedrofvteixeira:
…from an "unavailable" state to a state with data


pedrofvteixeira:
@pamval , @carlosrusso

PR #549, opened 2015-07-17
[CDF-600] Filter Component: 'Show "Only" Button' property is not work…

pedrofvteixeira:
…ing properly


pedrofvteixeira:
@pamval , @carlosrusso

PR #552, opened 2015-07-22
[TEST]

afrjorge:
- Ant properties are immutable, set junit.haltonfailure and junit.haltonerror before loading subfloor.xml


afrjorge:
@pamval please review

PR #553, opened 2015-07-22
[BISERVER-12645] maintain backwards compatibility

afrjorge:


afrjorge:
@pamval please review.

common-ui is no longer using the **cdf-legacy** dir (https://github.com/pentaho/pentaho-platform-plugin-common-ui/commit/df92977b3607d7b43cca1d8a801042f033f436b5#diff-b67f3bd8b98d11c1b1b0cc2a53412b1fR8). The same happens with plugin-reporting (https://github.com/pentaho/pentaho-platform-plugin-reporting/commit/9b230d1fdec517ad4288e47f82ea1ce9fa40337c#diff-ee51f17a85f879f2926547b577c0b1d1R12). So, although they are using the deprecated **content** endpoint no change is needed because the **js** folder has always existed as a **static-path** in all versions of CDF's plugin.xml.

PR #554, opened 2015-07-23
[CDE-582] Table component does not respect lifecycle.silent tag and still runs the block function

afrjorge:


afrjorge:
@pamval please review

pamval:
backport to 5.x, please

PR #558, opened 2015-07-28
[CDF-608] - escapeHtml no longer vulnerable to XSS

jvelasques:


jvelasques:
@pamval please review

PR #563, opened 2015-08-03
[CDF-615] - allow right alignment within dataBar-column

khauser:
Now with correct commits I hope ;)


khauser:
This is how it looks like:

image

Header text-align must be changed independently

jvelasques:
Manually merged

PR #564, opened 2015-08-03
[CDF-613, CDF-614] Passing arguments object to the callback functions…

plagoa:
… and code refactoring

- included the arguments object to the success and failure callbacks;
- refactored the ButtonComponent code:
  - now uses 'enable()' instead of explicitly adding class to the button element;
  - using 'setLabel()' instead of setting the button text directly;
  - extracted private method that determines if the button is a jQueryUi button;
  - cleaned up some code to remove some ambiguity or confusion with the scope of 'this' in the callback functions;


plagoa:
@pamval, @carlosrusso please review

pamval:
backport to 5.x

PR #566, opened 2015-08-03
[TEST] - Fixed failing test

plagoa:
- the previous commit didn't solve the problem of the failing test, which is the fact that enable() is now called when the buttonComponent is constructed and not just after being clicked. Updated the unit test accordingly;


plagoa:

PR #569, opened 2015-08-05
[CDF-615] - Allowing databar addin alignment. Default calculations ar…

jvelasques:
…e made using percentages now.


jvelasques:
@pamval please review

PR #574, opened 2015-08-07
[CDF-619] - CDE xaction components doesn't work

jvelasques:


jvelasques:
@pamval please review

PR #578, opened 2015-08-13
[CDE-615] - ButtonComponent doesn't show label

jvelasques:
- Adding a getter getWcdfSettings, for CDE to override
- Making Button and Table components default style tied to the rendererType


jvelasques:
@pamval please review

PR #583, opened 2015-08-18
[CDF-625] - Making sure wcdfSettings exists

pamval:


pamval:
@afrjorge please review

PR #585, opened 2015-08-21
[CDE-620]

afrjorge:
- added a dashboard reference to the state object parameter of the accordion addin implementation function so it can listen to and trigger events in the dashboard


afrjorge:
@pamval please review

pamval:
backport to 5.x

PR #589, opened 2015-08-25
[CDE-326] - MultiButton component: if the same button is clicked N time, processChange is called N times

jvelasques:


jvelasques:
@diogofscmariano please review, cherry-picking CDE-326 to 5.x

PR #597, opened 2015-09-10
[BACKLOG-5084] - Will not draw if placeholder not in DOM. This avoids…

jvelasques:
… setting a possible default parameter undefined.


jvelasques:
@pamval please review

krivera-pentaho:
I tested this locally and it works well.

PR #604, opened 2015-09-17
[CDF-626] - AutoCompleteBoxComponent will now update associated param…

jvelasques:
…eter as intended


jvelasques:
@pamval please review

PR #608, opened 2015-09-24
[CDF-349, CDF-351, DOC, TEST]

afrjorge:


afrjorge:
@pamval please review

afrjorge:
@pamval made changes according to your comments. Also, changed the return type of addComponent. Instead of returning a boolean value, if the component is added the dashboard object is returned allowing for chainable dashboard function calls ( e.g. dashboard.addComponent(comp).init() ), if it fails to be added an exception is thrown in order to avoid silent fails.

PR #612, opened 2015-09-28
[BACKLOG-5349] - Prevent potencial error if parameter value is null o…

diogofscmariano:
…r undefined. Added unit tests

@pamval please review


diogofscmariano:
dependent on https://github.com/pentaho/pentaho-platform-plugin-interactive-reporting/pull/278

pamval:
fix for legacy component should be added

PR #625, opened 2015-10-02
[CDF-637], [CDF-639], [CDF-659] and [CDF-660] 

mfgaspar:
Revamp to template component and add 9 addins to it
Revamp to the template addin and solve a bug

### DIFF
--- a/cdf-core/cdf/js-modules/addIns/colTypes/template/template.js
+++ b/cdf-core/cdf/js-modules/addIns/colTypes/template/template.js
@@ -15,10 +15,11 @@ define([
       '../../../AddIn',
       '../../../Dashboard',
       '../../../Logger',
+      '../../../dashboard/Utils',
       'amd!../../../lib/underscore',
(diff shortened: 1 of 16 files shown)


mfgaspar:
Pedro Value, I believe that this pull request should be in order.

pamval:
These addIns are very similar to the table ones. They should be the same code with some (if any) customizations for each of the use cases.

@afrjorge review and make the necessary changes: Maybe a TrendArrowBase that is inherited by a TrendArrow for table and a TrendArrow for Template.

mfgaspar:
[CDF-659] Revamp to the template addin: exposing a function for model parsing and also allow to customize messages as part of the options. Added to the pull request.

mfgaspar:
[CDF-660] - Solved a bug to the template addin. The addin was not applying the formatters in the correct way and might produce an invalid template.

afrjorge:
Please view https://github.com/webdetails/cdf/pull/633

PR #627, opened 2015-10-02
[CDE-558] Resultset created by postChange() should be persisted

afrjorge:


afrjorge:
@pamval please review, the data processed by a component's postFetch() is now persisted and available using component.query.lastProcessedResults() and component.query.getOption("lastProcessedResultSet")

PR #631, opened 2015-10-13
[CDF-572] - Adding method normalizeId to be overriden in embedded sce…

jvelasques:
…narios


jvelasques:
@pamval please review

PR #634, opened 2015-10-13
[CDF-659]

afrjorge:
- fallback to the message provided if it doesn't have a predefined value


afrjorge:
@pamval please review

PR #635, opened 2015-10-14
[CDF-609] - FilterComponent will now search server side with no pagin…

jvelasques:
…ation.

### DIFF
--- a/cdf-core/cdf/js-modules/components/filter/FilterComponent.js
+++ b/cdf-core/cdf/js-modules/components/filter/FilterComponent.js
@@ -182,7 +182,7 @@ define([
             /*
              * Handle empty datasets
              */
-            if (!configuration.component.search.serverSide && this.query.getOption('pageSize') === 0) {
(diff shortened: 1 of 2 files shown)


jvelasques:
@carlosrusso @pamval please review

jvelasques:
@pamval @carlosrusso added another commit

cdf-core/cdf/js-modules/components/filter/FilterComponent.js
```diff
@@ -182,7 +182,7 @@ define([
             /*
              * Handle empty datasets
              */
-            if (this.query.getOption('pageSize') === 0) {
+            if (!this.query || (!configuration.component.search.serverSide && this.query.getOption('pageSize') === 0)) {
```
carlosrusso: the argument of this if is very long. break the components into variables on their own:
var hasQuery = !! this.query;
var isPaginated = ....

PR #636, opened 2015-10-15
[CDF-609][CDF-641] - FilterComponent will now search server side. Accounting for when the query object does not exist.

jvelasques:


jvelasques:

pamval:
Aren't there tests covering this behavior ?

carlosrusso:
The component doesn't need a query to run, it can use a .valuesArray, so the query object might not exist.

jvelasques:
@pamval @carlosrusso @afrjorge, added unit tests, please review

pamval:
backport to 5.x

PR #641, opened 2015-10-16
[CDF-661] - Making blockG css FilterComponent specific

jvelasques:


jvelasques:
@pamval please review

PR #658, opened 2015-10-23
[CDF-663] - Search Term will no longer be sent to the server as lower…

jvelasques:
…case only


jvelasques:

pamval:
backport to 5.x,

PR #661, opened 2015-10-27
[CDF-667] - Adding root id option to a component with valuesArray

jvelasques:


jvelasques:

carlosrusso:
Please move the test to options.root.id to the updateModel method.
Also, please make the code a bit more idiomatic (boy scout rule), as the code still shows symptoms from the compilation from coffeescript

pamval:
really ? You're complaining about the code that the tool **you chose by yourself and used** generates ?
And telling someone else to fix it ?

jvelasques:
@pamval @carlosrusso changed accordingly, please review

carlosrusso:
@pamval Yes, indeed I am complaining about my own mistakes. All I have in mind is the quality of the product.  Should I refrain from doing so?

PR #662, opened 2015-10-28
[CDF-586] - FilterComponent can now receive a setting expandMode, to …

jvelasques:
…define if the expand div will push other page content or not


jvelasques:

carlosrusso:
seems fine to me

PR #663, opened 2015-10-29
[CDF-677] SortByValue incorrect results

afrjorge:
- when only one Sorting Addin is being used don't convert the resulting value into a String


afrjorge:
@pamval please review, the sorted values are only appended into the DOM so the test will be done using selenium

PR #667, opened 2015-10-30
[CDF-583] Refactor CDF's FilterComponent 'templates.js' to break it down into a set of moustache template files

afrjorge:
- removing some extra spaces that change the visual aspect of the UI when using the template HTML file


afrjorge:
@pamval please review

PR #672, opened 2015-11-10
[CDF-688] FilterComponent - filter.css is not the same in the legacy and amd versions of the component

afrjorge:
- made legacy and amd versions of filter.css equal, according to the UX team's guidelines
    - made template.css body rules more specific so AMD samples don't have overridden styles (e.g. font-family overridden by screen.css in a blueprint dashboard)


afrjorge:
@pamval please review

PR #673, opened 2015-11-11
[BISERVER-12840] - Accented Characters get encoded on Dropdown parameter ReportViewer

IvanNikolaychuk:
- Check if inputs are allready escaped and not apply re-escaping to them.
- Added tests

@pamval, @dcleao, review it please


pamval:
@jvelasques extending your previous negative lookahead solution to an extended set of chars. please review

jvelasques:
@IvanNikolaychuk @pamval this won't match &thetasym; -> the second rule should match until 8 characters - {2,8}
Looks good otherwise, far more complete than my previous approach :)

IvanNikolaychuk:
@jvelasques @pamval Quite right, updated

PR #677, opened 2015-11-12
[CDE-666] - Removing unnecessary extends, reverting some addin API ch…

jvelasques:
…anges for consistency


jvelasques:
@pamval please review

PR #683, opened 2015-11-12
[CDF-701] Split Filter Component's model into a base class and derivative class.

carlosrusso:
Moved some classes to the 'cdf/lib/' AMD namespace.
Also fixed [CDF-686]


carlosrusso:
@pamval please review.

pamval:
Make sure you mark the jira cases as resolved-

PR #686, opened 2015-11-13
[CDF-690] FilterComponent - checkbox out of position when using showGroupSelection on groups

afrjorge:


afrjorge:
@pamval please review, applied changes after discussing them with the UX team

PR #688, opened 2015-11-13
[TEST] Updated dependencies

carlosrusso:


carlosrusso:
@pamval this should solve the broken build

PR #691, opened 2015-11-16
[DOC] updated to AMD version of CDF

afrjorge:


afrjorge:
@pamval please review

PR #692, opened 2015-11-16
[CDF-701] Updated legacy version of the SelectionModel with the chang…

carlosrusso:
…es introduced in the require version


carlosrusso:
@pamval These changes are mostly needed by the legacy version of the NewMapComponent. Ran tests and built locally.

PR #694, opened 2015-11-20
[ENGOPS-1578] - Converting CDF to maven:

diogofscmariano:
- Fixing cdf-core js artefact reference in cdf-pentaho build


diogofscmariano:
@pamval @lgrill-pentaho please check
This will fix the cdf maven build

PR #696, opened 2015-11-20
[CDF-722] - Defaulting ajax requests to POST. Making sure i18n always runs with GET.

jvelasques:


jvelasques:
@pamval please review

PR #698, opened 2015-11-20
[CDF-718] - TableComponent now properly escapes Column Headers

jvelasques:


jvelasques:
@pamval please review

PR #702, opened 2015-11-23
[ENGOPS-1578] - Converting CDF to maven:

diogofscmariano:
- Fixing js artefacts publish by creating js modules for core, pentaho, and pentaho-cdf final artefact


diogofscmariano:
@lgrill-pentaho please review

PR #711, opened 2015-11-26
[CLEANUP] - Removing space from TableComponent css dependency

jvelasques:


jvelasques:
@pamval please review

PR #718, opened 2015-12-01
[CDE-714] - Adding the possibility to customize the path from which to load messages.properties files

jvelasques:


jvelasques:
@pamval please review

PR #719, opened 2015-12-01
[CDF-699] - CommentsComponent - only allowing saving when there is a comment

jvelasques:


jvelasques:
@pamval please review

PR #721, opened 2015-12-02
[CDF-456] jQuery Migrate plugin should be removed when no more deprecated features are detected

afrjorge:


afrjorge:
@pamval please review

PR #723, opened 2015-12-04
[CLEANUP, TEST]

afrjorge:
- removed duplicate error callback, added missing arguments parameter in CDF AMD version and updated tests
    - allow CDF legacy non-CI karma tests to execute multiple times
    - fixed path for cdf-base resource in non-CI legacy karma configuration file

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
+++ b/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
@@ -39,14 +39,15 @@
     deepProperties: ['defaults', 'interfaces'],
     defaults: {
       successCallback: function() {
-        Dashboards.log('Query callback not defined. Override.');
+        Dashboards.log('Query success callback not defined. Override.');
       },
-      errorCallback: function() {
-        if(Dashboards != undefined && Dashboards.handleServerError != undefined) {
-          Dashboards.handleServerError();
+      errorCallback: function(jqXHR, textStatus, errorThrown) {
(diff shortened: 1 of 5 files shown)


afrjorge:
@pamval please review

core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
```diff
@@ -39,12 +39,7 @@
     deepProperties: ['defaults', 'interfaces'],
     defaults: {
       successCallback: function() {
-        Dashboards.log('Query callback not defined. Override.');
-      },
-      errorCallback: function() {
```
pamval: Wouldn't it be safer to keep this one and get rid of the other one ?
That way we don't lose the defensive logic we have here.

afrjorge:
@pamval made changes according to your comments to keep the defensive logic when executing the default error callback and updated tests accordingly

core-js/src/test/javascript/config/karma.conf.legacy.js
```diff
@@ -112,7 +112,7 @@ module.exports = function(config) {
 
     // Continuous Integration mode
     // if true, it capture browsers, run tests and exit
-    singleRun: true,
+    singleRun: false,
```
pamval: is this intended ?

afrjorge: Yes, so we can edit/debug the specs and have the browser automatically reload them. This is for the non-CI configuration file only.

PR #728, opened 2015-12-16
[CDF-728] - Added jsdoc plugin to cdf build

diogofscmariano:
@pamval @annapopova1 @krivera @afrjorge @plagoa @jvelasques please check
This can be executed using mvn clean site


annapopova1:
@diogofscmariano you set "javascript.doc.dir" property in core-js/pom.xml but not in pentaho-js/pom.xml. Is it not necessary?

diogofscmariano:
This is ready to be merged. The issue is not related with the new plugin but it was related with the settings.xml changed.

PR #729, opened 2015-12-17
[CDF-787][CDF-788] Added a patch fixing OpenLayers.Layers.Google.v3

carlosrusso:


carlosrusso:
@pamval please review

PR #731, opened 2015-12-18
[CLEANUP] - Add README.md with build instructions

diogofscmariano:
@pamval @pmalves please review
I can do the same for cpf and cda once this gets accepted


diogofscmariano:
@pamval please recheck

PR #734, opened 2015-12-21
[CDF-456] Fix for backwards compatibility issue

afrjorge:
- replaced amd! loader plugin with shim configuration


afrjorge:
@pamval please review, using the amd! loader plugin was causing regression issues in 6.0

PR #736, opened 2015-12-22
[PPP-3455] - Dynamic code injection vulnerabilities found

akhayrutdinov:
- replace eval() with JSON.parse()

@diogofscmariano, @dcleao, @pamval, review it please.
I found the library on github: https://github.com/simile-widgets/timeline
Looks like it is abandoned, hence I suppose it makes sense to fix it locally.


pamval:
We need to add the original license to the header of these files (https://github.com/simile-widgets/timeline/blob/master/LICENSE.txt)
And then a comment saying we modified it.

akhayrutdinov:
@pamval, do you mean adding the header to all files or just putting the txt file into the folder?

pamval:
should be on the files :(

akhayrutdinov:
@pamval, I've done it in the separate commit

PR #755, opened 2016-01-20
[CDF-797] - Making sure escapeHtml only tries to escape strings, retu…

jvelasques:
…rning an empty string otherwise


jvelasques:
@pamval @diogofscmariano @afrjorge @plagoa @krivera-pentaho @annapopova1 please review

afrjorge:
Seems fine, the only usages of Utils.escapeHtml I found was in TableComponent and SelectBaseComponent, and the return value expected is always an escaped or empty string.

krivera-pentaho:
Looks good

PR #762, opened 2016-02-01
[BACKLOG-6242] - Added new template with initial version of the styli…

diogofscmariano:
…sation based on MindTouch current api style

@davidmsantos90 @afrjorge @pamval please check


diogofscmariano:
@dcleao just to keep you in the loop

PR #767, opened 2016-02-04
[TEST]

afrjorge:
- added missing dependencies
	- set included pattern object property to false for i18n *.properties files, no need to add <script> tags in the browser to load them (see http://karma-runner.github.io/0.13/config/files.html)


afrjorge:

PR #770, opened 2016-02-10
[CDF-827] - Added documentation for inner objects on Popup components

diogofscmariano:
[BACKLOG-6478] - Fixed template topics like:
 - Added new tag: code; to allow different case then example
 - Removed unneeded buttons for show/hide of Private, Inherited and Deprecated Class members
 - Fixed Method details template for Source Links
 - Removed duplicated code from member-details, using details (the correct one) instead
 - Updated method details for constructor, if that class is static and no constructor is needed
 - Updated Constructor location in the Class Page


diogofscmariano:
@afrjorge @pamval @jvelasques please review

PR #772, opened 2016-02-12
[BACKLOG-6345] WD: add component: duplicate component name 'PostInitMarker' error appear in browser's console while reopening dashboard

afrjorge:
- avoid adding the "PostInitMarker" component more than once


afrjorge:
@pentaho/orlandocalisbon please review

PR #794, opened 2016-02-17
[BACKLOG-6477] - Fixed the parent element of the namespace <ul> in the index page

graimundo:


graimundo:
@diogofscmariano please review

PR #801, opened 2016-02-22
[CDF - 809] - Samples under plugin-samples > CDF have some issues

eliofreitas:


eliofreitas:
samples updated, click without functions eliminated and solved errors present in the logs 
@pamval please review

PR #808, opened 2016-03-04
[CDF-787] Updated OpenLayers to Pentaho's patched version.

carlosrusso:
@pamval please review. The changes address [COMP-2429].


pamval:
ok, let's update the comp-2429 case with the new info and move on.

thanks

PR #812, opened 2016-03-07
[BACKLOG-6892] 6.1 Localization for PUC

ddiroma:
@mdamour1976


pamval:
@ddiroma I think I read somewhere that we're supposed to revert the translations to english. Is that correct ? Are we going to change them back at some point ?

ddiroma:
My understanding is we are supposed to revert some translations, but I have no idea if we are to change them back anytime.

PR #814, opened 2016-03-08
[CDF-871] PrptComponent sample executes the components in an arbitrary order

afrjorge:
- Set priority in Pentaho Reporting Component sample components. The select component must always execute before the reporting component.


afrjorge:
@pamval please review

PR #815, opened 2016-03-08
[CDF-872] Missing dependency in the CDF AMD broadcast sample

afrjorge:
- Added missing jquery dependency to the broadcast sample.


afrjorge:
@pamval please review

PR #826, opened 2016-03-14
[CDF-794] FilterComponent: Modified algorithm for determining if all/…

carlosrusso:
…no items are selected. (backport to 5.x)


carlosrusso:
@pamval if only all backports were this easy...

PR #827, opened 2016-03-14
[CDF-796] - On the Filter Component with the sortByLabel enabled, and…

carlosrusso:
… a page length defined: scrolling down to fetch more items yields an exception (backport to 5.x)


carlosrusso:
@pamval please review (and test)

PR #834, opened 2016-03-31
[CDF-875] - On the Filter Component with the sortByLabel enabled, and…

eliofreitas:
… a page length defined: scrolling down to fetch more items creates a loop until the last selected item is found


eliofreitas:

carlosrusso:
Seems fine

PR #835, opened 2016-04-05
[CDF-670] - As a dashboard developer using a TableComponent, I would …

eliofreitas:
…like to be able to provide a friendly error message when a query fails.


eliofreitas:
@jvelasques please review

PR #841, opened 2016-04-28
[BACKLOG-8006] - Table Component cannot updated if paginate Server si…

eliofreitas:
…de is true

### DIFF
--- a/core-js/src/main/javascript/cdf/queries/LegacyQuery.js
+++ b/core-js/src/main/javascript/cdf/queries/LegacyQuery.js
@@ -113,6 +113,9 @@ define([
             metadata: [makeMetadataElement(0)],
             resultset: json.values || []
           };
+          if(json.queryInfo) {
+            result.queryInfo = json.queryInfo;
+          }
           _.each(json.metadata, function(el, idx) {
             return result.metadata.push(makeMetadataElement(idx + 1, el));
           });
--- a/pentaho-js/src/test/javascript/cdf/components/TableComponent-spec.js
+++ b/pentaho-js/src/test/javascript/cdf/components/TableComponent-spec.js
@@ -196,5 +196,26 @@ define([
       dashboard.update(tableComponent);
     });
 
+    /**
+     * ## The Table Component # can update when server side is true
+     */
+    it("can update when server side is true", function(done) {
+      spyOn(tableComponent, 'update').and.callThrough();
+      spyOn($, 'ajax').and.callFake(function(params) {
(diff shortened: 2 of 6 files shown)


core-js/src/main/javascript/cdf/queries/LegacyQuery.js
```diff
@@ -111,7 +111,10 @@ define([
 
           var result = {
             metadata: [makeMetadataElement(0)],
-            resultset: json.values || []
+            resultset: json.values || [],
```
afrjorge: Is this necessary?

pentaho-js/src/test/javascript/cdf/components/TableComponent-spec.js
```diff
@@ -196,5 +196,27 @@ define([
       dashboard.update(tableComponent);
     });
 
+    /**
+     * ## The Table Component # can update when server side is true
+     */
+    it("can update when server side is true", function(done) {
```
krivera-pentaho: This test feels far more like a functional test than a unit test, since you have so many "callThroughs" and binding to the "cdf:postExecution". You should be making more use of the mock objects to constrain the bounds of the test.

diogofscmariano: This component is way too big and we need to test this following the cdf workflow. This is why we need these callThrough

eliofreitas: In addition this test takes into account all the component if for any reason another bug does not let the component render when the flag is up  (this problem happened before) this will warn us, if a more "specific" test is made we can be sure that the new code is fine but not necessarily the whole

krivera-pentaho: I am fine with making the test just test the new code. This will allow us to write more tests that are unit tests and eventually cover the whole component. I would prefer to do this instead of perpetuating functional unit tests incorrectly.

afrjorge: I think in this case we can keep the functional test added for the TableComponent and also add the unit tests, that are **missing**, to cover the changes in BaseQuery and LegacyQuery.

eliofreitas: am fine with making another unit test and letting this exist

krivera-pentaho: I am ok with this as well

eliofreitas:
updated

eliofreitas:
updated

PR #843, opened 2016-04-29
[BACKLOG-8006] -  Table Component cannot updated if paginate Server side

eliofreitas:
- fix for legacy queries (non-CDA)


afrjorge:
Seems fine. The CDF Table Component sample (using a legacy MDX query) is working with server side pagination, although these legacy queries lack the metadata information to support this feature.

PR #852, opened 2016-07-12
[CLEANUP] removed unnecessary console.log from specs

afrjorge:


afrjorge:
@pamval please review

PR #855, opened 2016-07-26
[PDB-1901] Filters used in Analyzer reports are not updated when using Dashboard buttons especially when the value of the buttons change

MikhailHubanau:
Added a new case to dashboard.fireChange() call condition in MultiButtonComponent.js#draw() function.

@pamval , @jvelasques please review.

### DIFF
--- a/core-js/src/main/javascript/cdf/components/MultiButtonComponent.js
+++ b/core-js/src/main/javascript/cdf/components/MultiButtonComponent.js
@@ -147,7 +147,7 @@ define([
           && myArray.length > 0) {
 
         //select first value
-        if((currentVal == null || currentVal == "" || (typeof(currentVal) == "object" && currentVal.length == 0))
+        if((currentVal == null || currentVal == "" || ((currentVal !== firstVal) && myArray.length == 1) || (typeof(currentVal) == "object" && currentVal.length == 0))
             && myself.parameter) {
 
           myself.dashboard.fireChange(myself.parameter, (myself.isMultiple) ? [firstVal] : firstVal);
--- a/core-js/src/test/javascript/cdf/components/MultiButtonComponent-spec.js
+++ b/core-js/src/test/javascript/cdf/components/MultiButtonComponent-spec.js
@@ -134,5 +134,143 @@ define([
 
       dashboard.update(multiButtonComponent);
     });
+
+    /**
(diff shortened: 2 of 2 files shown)


pamval:
@diogofscmariano @annapopova1 @afrjorge please review

diogofscmariano:
So, this is a case were we are drawing the buttons and in the same moment the first value is different from the one selected? Will this not break the use cases where buttons can have a different default value? Also, maybe it will better to use the !== instead !=. 

Can we have unit tests for this?

MikhailHubanau:
>So, this is a case were we are drawing the buttons and in the same moment the first value is different from the one selected? Will this not break the use cases where buttons can have a different default value? Also, maybe it will better to use the !== instead !=.

I updated PR. @diogofscmariano please take a look.

>Can we have unit tests for this?

No.

pamval:
Why can't we have unit tests?

MikhailHubanau:
I mean that for now we do not have MultiButtonComponent.js#draw() unit test.

pamval:
Have you taken a look at https://github.com/webdetails/cdf/blob/master/core-js/src/test/javascript/cdf/components/MultiButtonComponent-spec.js ?

I don't understand why you can't add an unit test there that covers the behavior you've introduced.
Any call to component.update will lead up to draw being called.

MikhailHubanau:
>Have you taken a look at https://github.com/webdetails/cdf/blob/master/core-js/src/test/javascript/cdf/components/MultiButtonComponent-spec.js ?

Yes.

>I don't understand why you can't add an unit test there that covers the behavior you've introduced.
Any call to component.update will lead up to draw being called.

I am trying to understand how can I check the count of myself.dashboard.fireChange() calls in two cases: myArray.length == 1 and myArray.length != 1, based on current MultiButtonComponent-spec.js code. I accepted that it will looks like 
//init
...
// execute
multiButtonComponent.draw()
//check
...
but test goes through dashboard.update(), I can not get this test mechanism for current moment.

MikhailHubanau:
Added unit test.

pamval:
@diogofscmariano @annapopova1 @afrjorge please review

MikhailHubanau:
@diogofscmariano @annapopova1 @afrjorge please review

PR #858, opened 2016-08-09

mtseu:
An impersonation guard has been implemented allowing users with admin role to work on behalf of other users.
The logic is encapsulated in the new ImpersonationHandler class which is called from StorageApi.impersonate() method.
In order to enable testability new class StorageEngineInterface is introduced and StorageEngine.setMockInstance() method
is added, which allows to provide mock for StorageEngine (the same way as this is implemented in
org.pentaho.platform.engine.security.SecurityHelper)


mtseu:
@pamval @diogofscmariano @jvelasques could you please review/merge?

PR #859, opened 2016-08-11

mtseu:
Re-enable @Ignored tests.


mtseu:
@pamval @diogofscmariano @jvelasques the tests had been annotated @Ignore while developing more tests and were forgotten to be re-enabled.

PR #862, opened 2016-08-23
[SP-2846] - Backport of CDF-896 - CDF Storage: any user is able to change the storage of another user (6.1 Suite)

mtseu:


mtseu:
@diogofscmariano @jvelasques @pamval 
I have no idea what is wrong with wingman. Could you please review/merge anyway?

core/src/main/java/org/pentaho/cdf/storage/StorageEngine.java
```diff
@@ -56,14 +57,22 @@ public static Operation get( String operation ) {
     }
   }
 
-  public static synchronized StorageEngine getInstance() {
+  public static synchronized StorageEngineInterface getInstance() {
```
mbatchelor: As my comment on the main case and the main PR, I don't believe this should be a public method. It's only here to facilitate testing, it should be package-protected not public.

mtseu: this method is called from classes in other packages:
https://github.com/webdetails/cdf/blob/master/pentaho/src/main/java/org/pentaho/cdf/context/ContextEngine.java#L397
https://github.com/mtseu/cdf/blob/master/pentaho/src/main/java/org/pentaho/cdf/render/CdfHtmlRenderer.java#L364

mtseu:
as per @mbatchelor comments - the way of mocking up the StorageEngine for unit tests has been reworked.
this pull request is squash and backport of https://github.com/webdetails/cdf/pull/863 and  https://github.com/webdetails/cdf/pull/858 

@mbatchelor I'd appreciate your review

PR #863, opened 2016-08-24

mtseu:
Functinality is still the same. Small refactor was done to leverage dependendancy injection for unit testing.


mtseu:
@mbatchelor as per our discussion I have reworked the way how to mock StorageEngine for unit-tests

PR #865, opened 2016-08-30
[PDB-1901] Filters used in Analyzer reports are not updated when usin…

mchen-len-son:
…g Dashboard buttons especially when the value of the buttons change

- copied the previous fix to cdf-legacy
- copied and updated unit tests

This PR is copying fix from #855 to cdf-legacy, also updating the unit tests. This is required in order to fix the issue in 5.4.


mchen-len-son:
@pamval @jvelasques @afrjorge please take a look.
thanks.

PR #866, opened 2016-09-01
[SP-2855] - Backport of PDB-1901 - Filters used in Analyzer reports a…

YuryBY:
…re not updated when using Dashboard buttons especially when the value of the buttons change (5.4 Suite)


YuryBY:
@jvelasques, @pamval, @mbatchelor, could you please take a look? - here is backport of https://github.com/webdetails/cdf/commit/fc72996af3d56dbde63b19a59c340cac0fa91233 to 5.4 since it use "legacy" dir.

YuryBY:
@dkincade adding to the thread

PR #872, opened 2016-09-30
[PPP-3577] - Removing Orient DB from CDF

pamval:
@diogofscmariano, this one.


pamval:
@diogofscmariano tested on the latest snapshot. this is good to be merged.

PR #873, opened 2016-10-03
[CDF-887] - Date Range input not able to pass parameter correctly to …

YuryBY:
…Analyzer component

@pamval, @jvelasques, could you please review the PR? Thanks.

### DIFF
--- a/pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
+++ b/pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
@@ -56,6 +56,12 @@
         $.map(this.parameters, function(k) {
             options[k[0]] = k.length == 3 ? k[2] : Dashboards.getParameterValue(k[1]);
         });
+        for (var option in options){
+            if (Date.parse(options[option])){
+                // see http://jira.pentaho.com/browse/CDF-887
+                options[option] = options[option] + " 00:00:00.0";
+            }
+        }
         return options;
     },
 
--- a/pentaho-js/src/main/javascript/cdf/components/AnalyzerComponent.js
+++ b/pentaho-js/src/main/javascript/cdf/components/AnalyzerComponent.js
@@ -57,6 +57,12 @@ define([
       $.map(this.parameters, function(k) {
         options[k[0]] = k.length == 3 ? k[2] : d.getParameterValue(k[1]);
       });
+      for (var option in options){
+          if (Date.parse(options[option])){
+              // see http://jira.pentaho.com/browse/CDF-887
+              options[option] = options[option] + " 00:00:00.0";
+          }
+      }
       return options;
     },
 


jvelasques:
Could/should we do this by having the Analyzer component set the property **dateFormat** on this component instead of changing this default?
If we do change the default, then [DateInputComponent](https://github.com/webdetails/cdf/blob/master/core-js/src/main/javascript/cdf/components/DateInputComponent.js) should also be changed, and [the associated test](https://github.com/webdetails/cdf/blob/master/core-js/src/test/javascript/cdf/components/DateInputComponent-spec.js) updated to cover this new default.
@pamval ^

pamval:
that would probably be a more self-contained fix. Otherwise, we'll change the default for every usage of the date component

YuryBY:
@pamval, @jvelasques, the PR is updated. Thanks.

pamval:
@YuryBY please update the require.js version of the component as well. Can't remember where that is, but @jvelasques knows

YuryBY:
@pamval, @jvelasques, have already done. Thanks.

pamval:
@jvelasques review and merge, please

jvelasques:
@YuryBY I tested this and its not fixing the issue...
I'll give it some more thought

YuryBY:
@jvelasques, I tested on snapshot build. It worked. Changes in input.js I tested on snapshot and nightly - worked everywhere.

jvelasques:
Not particularly fond of this solution, but can't think of anything else without changing the DateInputComponent also...
Is it possible to pass a parameter that evaluates as **Date** but was actually meant to be used as  something else? **String** or **Number**
The parameter passed might be a Number, which still parses correctly with **Date.parse()** for example...
We could restrict the condition even further:
`typeof options[option] !== "number" && !(options[option] instanceof Date)`
But I doubt I'm thinking of every possibility with this...
@pamval any thoughts?

YuryBY:
@jvelasques, @pamval, typeof dateFilter returns "string". If we will try something like !Number.parseInt(dateFilter)  it also does not work since returns 2016 for "2016-02-01"...

pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
```diff
@@ -56,6 +56,12 @@
         $.map(this.parameters, function(k) {
             options[k[0]] = k.length == 3 ? k[2] : Dashboards.getParameterValue(k[1]);
         });
+        for (var option in options){
+            if (Date.parse(options[option])){
+                // see http://jira.pentaho.com/browse/CDF-887
+                options[option] = options[option] + " 00:00:00.0";
```
mbatchelor: And what about DBs that don't report dates as timestamps ? Won't this fail ?

pentaho-js/src/main/javascript/cdf/components/AnalyzerComponent.js
```diff
@@ -57,6 +57,12 @@ define([
       $.map(this.parameters, function(k) {
         options[k[0]] = k.length == 3 ? k[2] : d.getParameterValue(k[1]);
       });
+      for (var option in options){
+          if (Date.parse(options[option])){
+              // see http://jira.pentaho.com/browse/CDF-887
+              options[option] = options[option] + " 00:00:00.0";
```
mbatchelor: And what about DBs that don't report dates as timestamps ? Won't this fail ?

pamval: @mbatchelor this pull request has been reverted and a new fix has been developed. The right pr is #880

YuryBY:
@mbatchelor, here is actual changes: https://github.com/webdetails/cdf/pull/880/files#diff-0fc8712fc13c08982fc1213e30e218aeR55 - now we have dateFormat as parameter

PR #874, opened 2016-10-03
[CDF-930] - Updating VisualizationAPI Component to work with VizApi 3.0

pamval:
@dcleao @afrjorge please review


pamval:
@jvelasques please review as well since the lazy @afrjorge is on vacation

pentaho-js/src/main/javascript/cdf/components/VisualizationAPIComponent.js
```diff
@@ -25,24 +30,49 @@ define([
...
+            createdModel.data = new Table(data);
+            createdModel.height= me.height;
+            createdModel.width = me.width;
+
+            $.each(me.vizOptions, function(i, v) {
+              var prop = createdModel.get(v[0]);
+              var value = me.dashboard.getParameterValue(v[1]);
 
-      visualSpec.type = this.vizId;
+              if (prop.type.isSimple) {
```
dcleao: This conversion scheme causes me chills :-D !
To be the most transparent as possible, we would require the values of parameters to be, directly, whatever the model accepts as inputs. This is the "spec" format (or direct instances, but that's not a common use case).
In the case of visual roles, unfortunately, this would require a value like: `{attributes: ["Country"]}` — it doesn't accept an array, directly.

If we keep this special syntax for visual roles, then, instead of using the `isSimple` test below, we'd be safer testing that the property actually represents a visual role. This way, other complex typed properties could still be used.

The way to test if a property is a visual role property is to test if its `type` attribute is a subtype of `pentaho/visual/role/mapping`:

```js
var mappingType = _context.get("pentaho/visual/role/mapping").type; // cache this, somewhere above.
if(!mappingType.isSubtypeOf(prop.type)) {
  createdModel.set(propName, value);
} else {
  createdModel.get(propName).attributes.set(value);
}
```

pentaho-js/src/main/javascript/cdf/components/VisualizationAPIComponent.js
```diff
@@ -25,24 +30,49 @@ define([
...
+      var me = this;
+      _context.getAsync(this.vizId)
+          .then( function (Model) {
+            var createdModel = new Model({});
+            createdModel.data = new Table(data);
+            createdModel.height= me.height;
+            createdModel.width = me.width;
+
+            $.each(me.vizOptions, function(i, v) {
+              var prop = createdModel.get(v[0]);
```
dcleao: This is not the "property type" instance, but the _current value_ of the property. If it is `null`, the line further down: `prop.type`, would throw... When it does not throw the obtained type is the type of the stored value, which can be a subtype of the declared property type.

To get the "property type" instance:

```js
var propType = createdModel.type.get(v[0]);

// The type of the value stored by properties of this type:
if(propType.type.isSimple) {
}
```

pentaho-js/src/main/javascript/cdf/components/VisualizationAPIComponent.js
```diff
@@ -25,24 +30,49 @@ define([
...
+              var value = me.dashboard.getParameterValue(v[1]);
 
-      visualSpec.type = this.vizId;
+              if (prop.type.isSimple) {
+                createdModel[prop] = value;
```
dcleao: Please do not use the JavaScript property getter/setter directly with generic code, cause these aren't defined when the name of the property is already taken in the class, for some other use. The generic safe way is to use the set (and get, for getting) method:
```js
createdModel.set(v[1], value);
```

pamval:
@dcleao partially reviewed, apart from having a single Model/view instance and the related (it seems to me) spec building in advance

PR #875, opened 2016-10-06
[CDF-931] - Logger.error not working properly when an invalid datasource is defined

kleysonr:


carlosrusso:
👍  @pamval this seems good to me

PR #877, opened 2016-10-11
[CDF-674] Restructuring the add-ins of the FilterComponent

carlosrusso:
A small intervention on the require version of the filter component, due to episodic OCD.

The goal here is to eventually kill the BaseFilter.Extensions namespace, which is not needed at all, and to remove the traces of the non-idiomatic code compiled from coffeescript.
Also, attempted to improve the code legibility around FilterComponent#getConfiguration

@pamval please review


pamval:
merge post code freeze

carlosrusso:
postponing PR

PR #881, opened 2016-10-27
[TEST] Fix JavaScript TypeError errors in core-js specs

afrjorge:


afrjorge:
@pamval @diogofscmariano please review, trying to minimize the amount of PhantomJS crashes in DEVCI

PR #882, opened 2016-10-27
[TEST] reviewed the usage of the optional "done" argument for async tests

afrjorge:
- `describe` blocks don't support the `done` parameter
- synchronous tests don't need the `done` parameter


afrjorge:
@diogofscmariano @pamval please review

PR #893, opened 2016-11-09
[CDE-849] Export of table with paginate server side set as true is only exporting 10 results in a kettle data source

Advent51:
fixed

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
+++ b/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
@@ -564,6 +564,12 @@
         queryDefinition.settingcolumnHeaders = options.columnHeaders;
       }
 
+      // only relevant for a component with server side pagination
+      if(options.exportPage === false) { // only export the entire dataset when exportPage is false
+        queryDefinition.pageSize = 0;
+        queryDefinition.pageStart = 0;
+      }
+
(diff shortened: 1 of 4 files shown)


pamval:
@jvelasques @afrjorge please review

core-js/src/main/javascript/cdf/queries/CdaQuery.js
```diff
@@ -212,6 +212,12 @@ define([
         queryDefinition.settingcolumnHeaders = options.columnHeaders;
       }
 
+      // only relevant for a component with server side pagination
```
afrjorge: please add the new `options.exportPage` parameter to the doc (e.g. in line 194)

core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
```diff
@@ -564,6 +564,12 @@
         queryDefinition.settingcolumnHeaders = options.columnHeaders;
       }
 
+      // only relevant for a component with server side pagination
+      if(options.exportPage !== true) { // only export a page explicitly
```
jvelasques: in order to preserve the default in all cases, change this condition to `=== false`
The comment can be updated accordingly `// only export the entire dataset when exportPage is false`

Advent51:
@afrjorge please review

PR #899, opened 2016-12-13
[BACKLOG-12494] - Include property to allow embeded dashboards

eliofreitas:


pamval:
why is one file commented and the other one not?

eliofreitas:
@diogofscmariano @pamval updated

PR #901, opened 2017-01-06
[CDF-949] CDF: Embeded Dashboards returns HTTP even when the connection is done through SSL/TLS (https)

AliaksandrShuhayeu:
[CDF-949] CDF: Embeded Dashboards returns HTTP even when the connection is done through SSL/TLS (https)
-added using scheme for HTTP protocol
-added tests
@pamval Could you please review and merge it?

### DIFF
--- a/pentaho/src/main/java/org/pentaho/cdf/CdfApi.java
+++ b/pentaho/src/main/java/org/pentaho/cdf/CdfApi.java
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2015 Webdetails, a Pentaho company. All rights reserved.
+ * Copyright 2002 - 2017 Webdetails, a Pentaho company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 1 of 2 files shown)


pamval:
@afrjorge @diogofscmariano please review

pentaho/src/main/java/org/pentaho/cdf/CdfApi.java
```diff
@@ -372,10 +373,13 @@ protected String getConfiguration( String path,
     return ContextEngine.getInstance().getConfig( path, parameterMap, inactiveInterval );
   }
 
-  protected String buildFullServerUrl( String protocol, String serverName, int serverPort  ) {
-    String p = "http";
+  protected String buildFullServerUrl( String protocol, String serverName, int serverPort, String scheme ) {
```
afrjorge: If your going with an extra parameter, I think using [isSecure](http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#isSecure()) is the best option because it also works if a load balancer is forwarding the requests using HTTP and has more support across different Java HTTP server implementations. Please see this discussion in [stackoverflow](http://stackoverflow.com/questions/8200853/how-can-i-know-if-the-request-to-the-servlet-was-executed-using-http-or-https).

AliaksandrShuhayeu:
@afrjorge  I have changed  code. Now it uses isSecure. Could you please review and merge it?

PR #914, opened 2017-03-09
[BACKLOG-14077] 

diogofscmariano:
@afrjorge @dcleao can you please review the js part?

### DIFF
--- a/core-js/src/main/javascript/cdf/components/ChartComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ChartComponent.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2015 Webdetails, a Pentaho company. All rights reserved.
+ * Copyright 2002 - 2017 Webdetails, a Pentaho company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
@@ -18,32 +18,34 @@ define([
   './CggComponent.ext',
   '../lib/jquery',
   'amd!../lib/underscore'
-], function(Utils, UnmanagedComponent, ChartComponentExt, CggComponentExt, $, _) {
+], function (Utils, UnmanagedComponent, ChartComponentExt, CggComponentExt, $, _) {
 
   return UnmanagedComponent.extend({
-    exportChart: function(outputType, overrides) {
+    exportChart: function (outputType, overrides) {
       var me = this;
 
-      var buildUrlParameters = function(overrides) {
+      var buildUrlParameters = function (overrides) {
         overrides = overrides || {};
 
         var urlParams = {};
 
         // Pass the parameters defined in this component to the used data source.
         var paramDefsArray = me.parameters;
-        if(paramDefsArray && paramDefsArray.length) {
+        if (paramDefsArray && paramDefsArray.length) {
           var paramDefs = $.extend({}, Utils.propertiesArrayToObject(paramDefsArray), overrides);
-          for(var name in paramDefs) {
-            if(paramDefs.hasOwnProperty(name)) {
+          for (var name in paramDefs) {
+            if (paramDefs.hasOwnProperty(name)) {
               // Works with eval ...
               var value = this.dashboard.getParameterValue(paramDefs[name]);
-              if($.isArray(value) && value.length == 1 && ('' + value[0]).indexOf(';') >= 0) {
+              if ($.isArray(value) && value.length == 1 && ('' + value[0]).indexOf(';') >= 0) {
                 // Special case where single element will wrongly be treated as a parsable array by cda
-                value = Utils.doCsvQuoting(value[0],';');
+                value = Utils.doCsvQuoting(value[0], ';');
               }
               //else Will not be correctly handled for functions that return arrays
 
-              if(typeof value == 'function') { value = value(); }
+              if (typeof value === 'function') {
+                value = value();
+              }
 
               urlParams['param' + name] = value;
             }
@@ -52,12 +54,12 @@ define([
 
         // Check debug level and pass as parameter
         var level = me.dashboard.debug;
-        if(level > 1) {
+        if (level > 1) {
           urlParams.paramdebug = true;
           urlParams.paramdebugLevel = level;
         }
 
-        var scriptName =  me.name.replace(/render_/, '');
+        var scriptName = me.name.replace(/render_/, '');
 
         urlParams.script = ChartComponentExt.getCccScriptPath(scriptName);
 
@@ -72,7 +74,7 @@ define([
       var url = CggComponentExt.getCggDrawUrl() + "?" + $.param(urlParams);
 
       var $exportIFrame = $('#cccExportIFrame');
-      if(!$exportIFrame.length) {
+      if (!$exportIFrame.length) {
         $exportIFrame = $('<iframe id="cccExportIFrame" style="display:none">');
         $exportIFrame[0].src = url;
         $exportIFrame.appendTo($('body'));
@@ -81,16 +83,18 @@ define([
       }
     },
 
-    renderChart: function() {
+    renderChart: function () {
       var cd = this.chartDefinition;
 
-      if(this.dashboard.isValidQueryDefinition(cd)) {
-        this.triggerQuery(cd, _.bind(this.render, this));
-      } else if(this.valuesArray != undefined) {
-        this.synchronous(_.bind(function() { this.render(this.valuesArray); }, this));
+      if (this.dashboard.isValidQueryDefinition(cd)) {
+        this.beginQuery(cd, _.bind(this.render, this));
+      } else if (this.valuesArray != undefined) {
+        this.execute(_.bind(function () {
+          this.render(this.valuesArray);
+        }, this));
       } else {
         // initialize the component only
-        this.synchronous(_.bind(this.render, this));
+        this.execute(_.bind(this.render, this));
       }
     }
   });
--- a/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2015 Webdetails, a Pentaho company. All rights reserved.
+ * Copyright 2002 - 2017 Webdetails, a Pentaho company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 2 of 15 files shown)


core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -77,21 +97,109 @@ define([
...
+        if (cd.stacked == true || fullName == "area") {
+          fullName += "Stacked";
+        }
+      }
+
+      return fullName;
+    },
+
     render: function(values) {
+      var myself = this;
```
dcleao: On another end, I would not remove the base class' `renderChart` method: there are dashboards out there overriding this (not very orthodox, I know) and I think there might be other ways to handle this.

Here's an alternative that I think would work better:

```js
var chart = {
  renderChart: function() {		
    
    // Always executes asynchronously.

    var cd = this.chartDefinition;		
    
    if(this.dashboard.isValidQueryDefinition(cd)) {
      
      this.beginQuery(cd, _.bind(this._renderOuter, this));

    } else if(this.valuesArray != null) {

      this.execute(_.bind(function() { this._renderOuter(this.valuesArray); }, this));

    } else {
      // Initialize the component only
      this.execute(_.bind(this._renderOuter, this));
    }
  },

  _shouldUseVizApiStyles: function() {
    return !!this._vizApiStyles && !_.contains(chartTypesBlackList, this.getCccChartName());
  },

  _renderOuter: function(data) {

    var promiseExtensions;
       
    this._preProcessChartDefinition();

    if(this._shouldUseVizApiStyles()) {

      promiseExtensions = context.getAsync("pentaho/ccc/visual/" + this.getCccChartName())
          .then(function(View) {
             return View.type.extensionEffective;
          });
    } else {
      // no external extensions
      promiseExtensions = Promise.resolve(null);
    }

    promiseExtensions
      .then(_.bind(this._renderInner, this, data))
      .then(_.bind(this.endExec, this), _.bind(this.failExec, this));
  },

  render: function(data) {
    this._renderInner(data, null);
  },

  // Similar to yours `_internalRender` and to the old "render" method.
  _renderInner: function(data, externalChartDefinition) {
    
    $("#" + myself.htmlObject).append('<div id="' + myself.htmlObject + 'protovis"></div>');

    // Always clone the original chartDefinition.
    var cd = $.extend({}, this.chartDefinition);

    if(externalChartDefinition) cd = $.extend(externalChartDefinition, cd);

    cd.canvas = this.htmlObject + "protovis";

    // Process extension points
    if(cd.extensionPoints) {
      var ep = {};
      cd.extensionPoints.forEach(function(a) { ep[a[0]] = a[1]; });
      cd.extensionPoints = ep;
    }

    if(this._shouldUseVizApiStyles() && cd.colors == null) {
      cd.colors = paletteRegistry.get("cdf_default").colors;
    }

    this.chart = new this.cccType(cd);

    if(arguments.length > 0) {
      this.chart.setData(data, {crosstabMode: this.crosstabMode, seriesInRows: this.seriesInRows});
    }

    this.chart.render();
  }
}
```

core-js/src/main/javascript/cdf/components/ccc/vizApi.conf.js
```diff
@@ -0,0 +1,646 @@
...
+            format: {
+              number: {
+                style: numberStyle
+              }
+            },
+
+            errorMessage_visible: false,
+            noDataMessage_visible: false,
+            invalidDataMessage_visible: false,
+
```
dcleao: These are probably not good defaults for CDF

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -77,21 +97,109 @@ define([
...
+        myself._internalRender.apply(myself, [values, View.type.extensionEffective]);
+      }).then(_.bind(myself.endExec, myself), _.bind(myself.failExec, myself));
+    },
 
-      var o = $.extend({}, this.chartDefinition);
+    /**
+     *
+     */
+    _internalRender: function(values, externalChartDefinition){
+      var o = externalChartDefinition ? $.extend(externalChartDefinition, this.chartDefinition) : this.chartDefinition;
```
dcleao: Still not sure that simply overwriting whatever the user might have specified in `preProcess` and `postFetch` is the way to go.

core-js/src/main/javascript/cdf/components/ccc/vizApi.conf.js
```diff
@@ -0,0 +1,646 @@
...
+            plotFrameVisible: false,
+            plotSizeMin: {width: 600, height: 300},
+
+            clickable:  false,
+            selectable: false,
+            hoverable:  true,
```
dcleao: `clickable` should be removed.
Not sure if `selectable` should be set to `false` — only if some of the styling changes we did don't work well with selection...

diogofscmariano:
@dcleao can you please recheck? Still need to add unit tests and doc

core-js/src/main/javascript/cdf/components/ProtovisComponent.js
```diff
@@ -23,11 +23,11 @@ define([
       if(this.parameters == undefined) {
         this.parameters = [];
       }
-
+      
       this.renderChart();
     },
 
-    render: function(values) {
+    _renderOuter: function(values) {
```
dcleao: If you see my comment below, maybe we can just call `render` to `_renderOuter` and not need both. So this rename here would not be needed.

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -14,52 +14,147 @@
...
+  var _stackedExceptions = ['barNormalized'];
 
   return ChartComponent.extend({
 
     query: null,
     chart: null,
 
-    _preProcessChartDefinition: function() {
+    _cccChartName: null,
+    getCccChartName: function () {
```
dcleao: Still think this name is not appropriate. Did you see my suggestion in the previous review?

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -14,52 +14,147 @@
...
+    /**
+     *
+     * @param name
+     * @returns {*}
+     * @private
+     */
+    _getCccFullName: function (name) {
+      var fullName;
+      //get the ccc name first
+      var m = name.split(_reCccName);
```
dcleao: Use of split to do this doesn't seem right — not the right tool for the job, even if you can get it to do what you want. For this use case, split returns an array of 4 entries...

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -68,49 +163,81 @@ define([
...
-        pv.listenForPageLoad(function() {
+        pv.listenForPageLoad(function () {
           me.renderChart();
         });
       }
     },
 
-    render: function(values) {
+    render: function (data) {
+      this._renderOuter(data);
```
dcleao: I might have complicated thing a bit, while trying to keep `render` backwards compatible... I think we can just live with renaming the current `_renderOuter` to the old name `render`. No? The difference will be semantic. While, before, `render` was itself synchronous and did not finish the execution, now, it is asynchronous and will finish the execution...

core-js/src/main/javascript/cdf/components/ChartComponent.js
```diff
@@ -84,13 +84,15 @@ define([
     renderChart: function() {
       var cd = this.chartDefinition;
 
-      if(this.dashboard.isValidQueryDefinition(cd)) {
-        this.triggerQuery(cd, _.bind(this.render, this));
-      } else if(this.valuesArray != undefined) {
-        this.synchronous(_.bind(function() { this.render(this.valuesArray); }, this));
+      if (this.dashboard.isValidQueryDefinition(cd)) {
+        this.beginQuery(cd, _.bind(this._renderOuter, this));
```
dcleao: Keep calling the `render` method here (see comments below).

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -68,49 +163,81 @@ define([
...
+    
+    _renderInner: function (data, externalChartDefinition) {
       $("#" + this.htmlObject).append('<div id="' + this.htmlObject + 'protovis"></div>');
 
-      this._preProcessChartDefinition();
+      // Always clone the original chartDefinition.
+      var cd = $.extend({}, this.chartDefinition);
+
+      if (externalChartDefinition) {
+        cd = $.extend(externalChartDefinition, cd);
```
dcleao: I know I wrote this bit... but it is intended that chartDefinition prevails right? `externalChartDefinition` are default values. Just checking.

core-js/src/main/javascript/cdf/components/ccc/vizApi.conf.js
```diff
@@ -0,0 +1,638 @@
...
+
+            // . dot
+            dot_lineWidth: 0,
+            dot_fillStyle: function() {
+              var c = this.delegate();
+              return this.finished(c && !this.scene.isActive ? c.alpha(0.5) : c);
+            }
+          },
+
+          props: {
```
dcleao: `props` are of the `model`, not the `view`. Please remove.

core-js/src/main/javascript/cdf/components/ccc/vizApi.conf.js
```diff
@@ -0,0 +1,638 @@
...
+            bar_fillStyle: function() {
+              var c = this.delegate();
+              return this.finished(c && this.scene.isActive ? c.alpha(0.5) : c);
+            }
+          }
+        }
+      },
+
+      // [Visual Roles] - Bars, Columns, Line, Area
+      {
```
dcleao: `props` are of the `model`, not the `view`. The whole rule can be removed.

core-js/src/main/javascript/cdf/components/ccc/vizApi.conf.js
```diff
@@ -0,0 +1,638 @@
...
+
+      // Heat Grid
+      {
+        select: {
+          priority: -1,
+          application: "pentaho-cdf",
+          type: "pentaho/ccc/visual/heatGrid"
+        },
+        apply: {
+          props: {
```
dcleao: `props` are of the `model`, not the `view`. Please remove.

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
diogofscmariano: Yes. The objective is to have chartDefinition prevailing. The idea is that whatever the dev does with its component (and we at compat 3 mode), the moment we get into the render the viz api definitions will be smashed into the chart definitions, not overriding what was there. In parallel with this, the proposal is to remove some ccc property defaults to allow that when the devs dont change their property (either in CDE or CDF) the Viz Api definitions come in. Does this makes sense?

core-js/src/main/javascript/cdf/components/ChartComponent.js
```diff
@@ -18,32 +18,34 @@ define([
...
-              if($.isArray(value) && value.length == 1 && ('' + value[0]).indexOf(';') >= 0) {
+              if ($.isArray(value) && value.length == 1 && ('' + value[0]).indexOf(';') >= 0) {
                 // Special case where single element will wrongly be treated as a parsable array by cda
-                value = Utils.doCsvQuoting(value[0],';');
+                value = Utils.doCsvQuoting(value[0], ';');
               }
               //else Will not be correctly handled for functions that return arrays
 
-              if(typeof value == 'function') { value = value(); }
+              if (typeof value == 'function') {
```
afrjorge: please change to `===` or `_.isUndefined(value)`

dcleao: I vote on `===`.

pentaho-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.ext.js
```diff
@@ -0,0 +1,168 @@
...
+   * @private
+   */
+  var _reCccName = /^(.*)Chart$/;
+
+  var _context = PentahoTypeContext.getInstance();
+
+  /**
+   * List of viz types to exclude from the Viz Api chart defaults
+   *
+   * @type {[String]}
```
dcleao: `{string[]}` or `Array.<string>`

pentaho-js/src/test/javascript/cdf/components/ccc/BaseCccComponent.ext-spec.js
```diff
@@ -0,0 +1,127 @@
...
+ */
+
+define([
+  'cdf/components/ccc/BaseCccComponent.ext'
+], function (BaseCccComponentExt) {
+
+  /**
+   * ## The Base Ccc Component Ext
+   */
+  fdescribe("The Base Ccc Component Ext #", function () {
```
dcleao: forgotten `fdescribe`.

diogofscmariano:
@afrjorge @dcleao pr updated
i will check the cgg case in another pr. But it should apply the viz api styles

PR #917, opened 2017-03-20
[BACKLOG-14077] 

diogofscmariano:
- Added new chart default font

@afrjorge @dcleao @pamval please review
We should not need any component request because this font is already used by det. Just adding it here to make sure the chart defaults declared in the viz api related with fonts can be properly applied


diogofscmariano:
dont merge yet, new changes arriving

diogofscmariano:
@dcleao @afrjorge  new change added, can you review?

PR #921, opened 2017-03-23
[CDA-211]-Implement a whitelist for CORS

EgorZhuk:
-the test is moved to CPF

@pamval @mchen-len-son @mbatchelor please review


EgorZhuk:
@diogofscmariano @carlosrusso @pamval @mbatchelor  please review

PR #923, opened 2017-03-27
[BACKLOG-14077]

diogofscmariano:
- Change viz type blacklist to accommodate new types added to common-ui. Tests updated accordingly
 - Fixed bug were baseAxisLabelDesiredAngles and colors were not being correctly overridable

@dcleao please review

### DIFF
--- a/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
@@ -144,13 +144,20 @@ define([
       // Always clone the original chartDefinition.
       var cd = $.extend({}, this.chartDefinition);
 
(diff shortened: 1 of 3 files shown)


pentaho-js/src/main/javascript/cdf/components/ccc/config/cdf.vizApi.conf.js
```diff
@@ -21,6 +21,22 @@ define([], function() {
         priority: -1,
         select: {
           application: "pentaho-cdf",
+          type: [
+            "pentaho/ccc/visual/heatgrid",
```
dcleao: This margin for heatgrid is not specific to cdf. We'll do it in common-ui instead.

PR #928, opened 2017-04-19
[BACKLOG-14318] Sync with common-ui template upgrade

davidmsantos90:
@dcleao @carlosrusso @graimundo @nantunes @marcovala @diogofscmariano please review.

Sync with the upgrade done in the https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1031 template . Tested and the documentation was generated correctly

### DIFF
--- a/core-js/pom.xml
+++ b/core-js/pom.xml
@@ -12,6 +12,7 @@
   <packaging>pom</packaging>
   <properties>
     <js.project.list>ccc</js.project.list>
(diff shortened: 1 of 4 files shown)


davidmsantos90:
@diogofscmariano @afrjorge @eliofreitas please review. I tested the output of running the new template, and regarding what I changed (types links and source links) were giving the same output as before

PR #929, opened 2017-05-22
[CDE-906] - this was window when the setTimeout was being called

pamval:


pamval:
@diogofscmariano please review

PR #931, opened 2017-06-21
[BISERVER-13677]: Report with Post Processing Formula that references…

TatsianaKasiankova:
… another parameter goes into a loop and it never stops

Added moment-timezone-with-data.js library to work with dates in any timezone.
It contains all the zone data right up to Version 2017b.


TatsianaKasiankova:
@mbatchelor @pamval @mchen-len-son @afrjorge  please review

PR #939, opened 2017-07-20
[SP-3677]: Backport of BISERVER-13677 - Report with Post Processing F…

TatsianaKasiankova:
…ormula that references another parameter goes into a loop and it never stops (7.1 Suite)

Cherry-pick of the fix for [BISERVER-13677]: Report with Post Processing Formula that references another parameter goes into a loop and it never stops


TatsianaKasiankova:
@pamval , @mchen-len-son , @stepanovdg please review.

PR #940, opened 2017-07-20
[CDF-960] Dashboards fail to load sporadically due to an improper configuration of cdf/lib/sanitizer/* modules

afrjorge:


afrjorge:
@pamval @carlosrusso please review.
This issue is breaking builds and runtime dashboards randomly, depending on the order the AMD modules get loaded.

pamval:
Ammend the commit message to reference only one jira case.

afrjorge:
Commit message amended to reference only one Jira case.

PR #944, opened 2017-08-16
[PPP-3817] - Removal of some more samples and removed non existent de…

pamval:
…pendency from resources.properties


pamval:
@diogofscmariano please review and merge.

PR #945, opened 2017-08-18
[PPP-3817] - Still some more samples to be excluded

pamval:
Removed queries folder under 20-Samples. Legacy queries are no longer supported.


pamval:
@diogofscmariano can you merge this please?

PR #946, opened 2017-08-23
[CDF-985] Fixed the issue when not escaped html value for text …

annapopova1:
…input component breaks it.

For example, the value is '<>' and as a result we had html like <div id="46205-98607"><input type="text" id="46205-98607-input" name="46205-98607-input" value="" <="">''&gt;</div>

@afrjorge @pentaho-arua @eliofreitas @nunodelisboa please review

### DIFF
--- a/core-js/src/main/javascript/cdf/components/TextInputComponent.js
+++ b/core-js/src/main/javascript/cdf/components/TextInputComponent.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2015 Webdetails, a Pentaho company. All rights reserved.
+ * Copyright 2002 - 2017 Webdetails, a Pentaho company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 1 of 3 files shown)


core-js/src/main/javascript/cdf/components/TextInputComponent.js
```diff
@@ -20,30 +20,18 @@ define([
...
-      if(myself.maxLength) {
-        Logger.warn("Attribute 'maxLength' is deprecated");
-      }
-
-      myself.placeholder().html(selectHTML);
-
-      var el = $("#" + name);
+      myself._addHtmlToPlaceholder();
 
+      var el = $("#" + this.name);
```
nunodelisboa: Isn't this `this` supposed to be `myself`?

core-js/src/main/javascript/cdf/components/TextInputComponent.js
```diff
@@ -20,30 +20,18 @@ define([
...
+      el.val(this.dashboard.getParameterValue(this.parameter));
```
nunodelisboa: Arent't these `this` themselves supposed to be `myself` ?

core-js/src/main/javascript/cdf/components/TextInputComponent.js
annapopova1: in this line it's the same. myself makes sense for callbacks

core-js/src/main/javascript/cdf/components/TextInputComponent.js
annapopova1: the same as above

core-js/src/main/javascript/cdf/components/TextInputComponent.js
```diff
@@ -52,6 +40,19 @@ define([
 
     getValue: function() {
       return $("#" + this.name).val();
+    },
+
+    _addHtmlToPlaceholder: function() {
```
nunodelisboa: Isn't it easier to read if it is something like
```
function() {
  var attrs = { 
    type: "text",
    id: this.name,
    name: this.name,
    size: this.size || this.charWidth || undefined,
    maxlength: this.maxLength || this.maxCharts || undefined
};

var componentHTML = '<input ' + processAttrs(attrs) + '>';

if (this.size) Logger.warn("Attribute 'size' is deprecated");
if (this.maxLength) Logger.warn("Attribute 'maxLength' is deprecated");

this.placeholder().html(componentHTML);

function processAttrs(conf) {
  var list = [];
  for (prop in conf){
    if (conf.hasOwnProperty(prop). && conf[prop] != null){
      list.push(prop + '="' + conf[prop] + '"');
    }
  }
   return list.join(" ");
 }
}
```

nunodelisboa: Also, maybe use the `sanitiser` on `conf[prop]` just to be safe

PR #948, opened 2017-08-24
[BACKLOG-16943] Adding environment config to cdf-embed.js

davidmsantos90:
@webdetails/millenniumfalcon please review

 - CDE's embed js endpoint calls the same CDF endpoint, so those two cases are covered with this commit


dcleao:
When should `cdf-embed.js` start not including what's already included in `webcontext.js`?

PR #952, opened 2017-09-06
[CDF-986] XSS vulnerability in Date Input Component

Advent51:
- input value sets through jquery

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/components/input.js
+++ b/core-js/src/main/javascript/cdf-legacy/components/input.js
@@ -633,9 +633,12 @@ var DateInputComponent = BaseComponent.extend({
     //ToDo: stretch interval to catch defaultValue?..
     //Dashboards.getParameterValue(myself.parameter))
 
+    var $el = $('<input class="date-input" id="' + inputId + '"/>');
+    $el.val(inputValue);
+
     myself.placeholder()
-        .addClass('date-input-container')
-        .html('<input class="date-input" id="' + inputId + '" value="' + inputValue + '"/>');
+      .addClass('date-input-container')
+      .html($el);
 
     $(function(){
       myself.placeholder("input").datepicker({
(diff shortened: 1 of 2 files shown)


Advent51:
@afrjorge @annapopova1 @mchen-len-son @pamval please review

core-js/src/main/javascript/cdf-legacy/components/input.js
```diff
@@ -635,7 +635,9 @@ var DateInputComponent = BaseComponent.extend({
 
     myself.placeholder()
         .addClass('date-input-container')
-        .html('<input class="date-input" id="' + inputId + '" value="' + inputValue + '"/>');
+        .html('<input class="date-input" id="' + inputId + '"/>');
+
+      $('#' + inputId).val(inputValue);
```
annapopova1: looks like incorrect indent here, should be 4 spaces for this line.

Advent51: Thanks!
Now it should be fixed.

core-js/src/main/javascript/cdf/components/DateInputComponent.js
```diff
@@ -47,7 +47,9 @@ define([
 
       myself.placeholder()
         .addClass('date-input-container')
-        .html('<input class="date-input" id="' + inputId + '" value="' + inputValue + '"/>');
+        .html('<input class="date-input" id="' + inputId + '"/>');
+
+      $('#' + inputId).val(inputValue);
```
afrjorge: Using a selector to search for an element that is being created here seems to be excessive and might affect performance.

annapopova1: @afrjorge we can't do this in another way. And we do absolutely the same for text input component

afrjorge: I don't understand. Is there any reason for the following not to work?
```javascript
var $el = $('<input class="date-input" id="' + inputId + '"/>');
$el.val(inputValue);

myself.placeholder()
  .addClass('date-input-container')
  .html($el);
```

Advent51:
@mbatchelor @pamval please review

PR #954, opened 2017-09-20
[PPP-3797] Use of vulnerable component mustache.js

mchen-len-son:
- updated mustache.js to 2.2.1 version


mchen-len-son:
Checked library's release notes, checked mustache usages in code. No breaking changes are expected.
Replaced mustache.js in a local build - opened several CTools reports - no errors in browser console.

@pamval can you please take a look?

pamval:
@mchen-len-son what kind of tests were run here? 
We should run the automation tests for cdf to make sure we don't have regressions.

PR #972, opened 2017-11-09
[PDB-1987]-Check box filters with no default value set, display the first checkbox as selected

denisprotopopov:
-fixed

### DIFF
--- a/core-js/src/main/javascript/cdf/components/MultiButtonComponent.js
+++ b/core-js/src/main/javascript/cdf/components/MultiButtonComponent.js
@@ -143,6 +143,10 @@ define([
           if(!myself.isMultiple) { break; }
         }
       }
+      if (myself.useFirstValue !== undefined && myself.useFirstValue == false){
+        foundDefault = true;
+        myself.defaultIfEmpty = false;
+	  }
       if(((!foundDefault && !myself.isMultiple) || (!foundDefault && myself.isMultiple && myself.defaultIfEmpty))
           && myArray.length > 0) {
 
--- a/core-js/src/main/javascript/cdf/components/SelectBaseComponent.js
+++ b/core-js/src/main/javascript/cdf/components/SelectBaseComponent.js
@@ -136,6 +136,12 @@ define([
        */
       var isEmpty = currentVals == null;
       var hasChanged = !currentIsValid;
+	  
+	  if (this.useFirstValue !== undefined && this.useFirstValue == false){
+		  if (isEmpty){
+			  this.defaultIfEmpty = false;
+		  }
+	  }
       if(isEmpty && this.defaultIfEmpty && firstVal != null) {
         // Won't remain empty
         currentVals = [firstVal];
--- a/core-js/src/main/javascript/cdf/components/ToggleButtonBaseComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ToggleButtonBaseComponent.js
@@ -43,6 +43,11 @@ define(['../lib/jquery', './InputBaseComponent'], function($, InputBaseComponent
           }
         }
       }
+	  if (myself.useFirstValue !== undefined && myself.useFirstValue == false){
+		  if (!hasCurrentVal){
+			  myself.defaultIfEmpty = false;
+		  }
+	  }
       // if there will be no selected value, but we're to default if empty, select the first
       if(!hasCurrentVal && myself.defaultIfEmpty) {
         currentValArray = [myArray[0][vid]];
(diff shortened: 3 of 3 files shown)


pamval:
@afrjorge please review

denisprotopopov:
@afrjorge please review

denisprotopopov:
@afrjorge please review

afrjorge:
@denisprotopopov CDF framework components don't have a `useFirstValue` property, so what this code does is disable some functionalities, i.e. it's equal to adding `&& false` to those `if` statements. Plus, if you do want to add a new property you need to add documentation in the component samples and add support for the property in the CDE editor, so users can actually use it when building dashboards.
It seems that the issue that originated this is just a bad usage of a prompt to allow selecting the values shown in a table. If the prompt has the default value configured, and the table is correctly configured to read from it, the issue should not happen. But instead, the default value is being set in the table and thus the prompt doesn't have access to it while initialising.

denisprotopopov:
@afrjorge could you please clarify what correct behaviour should be in this case? Should be set EMEA value for prompt "Specify" field? Or shouldn't be allow set empty value for "Specify"?

afrjorge:
Please see comments in the Jira issue.

denisprotopopov:
@afrjorge please review

core-js/src/main/javascript/cdf/dashboard/Dashboard.lifecycle.js
```diff
@@ -638,6 +638,9 @@ define([
       }
 
       if(object.isManaged === false && object.update) {
+        if (object.defaultValue == null && !object.useFirstValue)  {
```
afrjorge: It might be better to not add component logic into the Dashboard sub-modules.

denisprotopopov: @afrjorge why not ? this logic needed in dashboard

afrjorge: But it's not dashboard behavior, it's component behavior. `object` is the component, and only 5 components use the `defaultIfEmpty` (`useFirstValue`), so it's not even general component behavior, it's specific to components that extend from the SelectBase or ToggleButtonBase components.

denisprotopopov: @afrjorge I've moved logic into components, please look at it

afrjorge: When a user creates a CDE dashboard he will not be able to, for example, add a select component and have the first value selected because, in this case, _defaultIfEmpty && useFirstValue_ (defaults _true && undefined_) is _falsy_, and because the properties are not editable in the CDE editor, selecting the first value will not be possible.

core-js/src/main/javascript/cdf/components/SelectBaseComponent.js
```diff
@@ -136,7 +136,7 @@ define([
        */
       var isEmpty = currentVals == null;
       var hasChanged = !currentIsValid;
-      if(isEmpty && this.defaultIfEmpty && firstVal != null) {
+      if(isEmpty && this.defaultIfEmpty && firstVal != null && this.useFirstValue == true) {
```
afrjorge: When a user creates a CDE dashboard he will not be able to, for example, add a select component and have the first value selected because, in this case, _defaultIfEmpty && useFirstValue_ (defaults _true && undefined_) is _falsy_, and because the properties are not editable in the CDE editor, selecting the first value will not be possible.
To do this, tests, documentation and support for the properties in the CDE editor is needed for each affected component.

denisprotopopov:
@afrjorge please review

denisprotopopov:

core-js/src/main/javascript/cdf/components/MultiButtonComponent.js
```diff
@@ -143,8 +143,14 @@ define([
           if(!myself.isMultiple) { break; }
         }
       }
+	  var useFirstValue = true;
```
afrjorge: _useFirstValue_ and _defaultIfEmpty_ serve the same purpose, we only need one variable to control the behavior of selecting the first value if none is selected. The same applies for the other components.

denisprotopopov:
sent another PRs;
https://github.com/webdetails/cdf/pull/974
https://github.com/webdetails/cde/pull/805
https://github.com/pentaho/pentaho-platform-plugin-common-ui/pull/1224
https://github.com/pentaho/pentaho-platform-plugin-dashboards/pull/436

PR #977, opened 2018-01-12
[CDF-972] Filter component doesn't support embedding with CORS

afrjorge:
- reviewed component minification


afrjorge:
@pamval please review, should be merged along with https://github.com/webdetails/cde/pull/815

PR #979, opened 2018-02-27
[BACKLOG-20459] Downstream updates

axelguiloff:
Depends on https://github.com/webdetails/ccc/pull/331

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,29 +32,32 @@
       <url>dav:${site.publish.url}/${project.groupId}/cdf-plugin-${project.version}</url>
     </site>
   </distributionManagement>
+
   <properties>
-    <dependency.hibernate.revision>3.2.6.ga</dependency.hibernate.revision>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -68,7 +68,7 @@
         <version>${dependency.cpf.revision}</version>
       </dependency>
       <dependency>
-        <groupId>pentaho</groupId>
+        <groupId>org.pentaho.ctools</groupId>
         <artifactId>ccc</artifactId>
         <version>${dependency.ccc.revision}</version>
```
afrjorge: You could also update the property names, here they still use the deprecated syntax *dependency.#.revision*, but should use *#.version*.
If I remember correctly, dependencies that are ours should use the property names contained in:
https://github.com/pentaho/build-resources/blob/8.1.0.0/version.properties
If for now you don't update CDF at least make sure this rule is enforced in CCC.

PR #982, opened 2018-03-12
[BACKLOG-22080] add flags for CDF to avoid components (table and CCC)…

joaormf:
… to flicker when automatically refreshed during realtime use-case

### DIFF
--- a/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
@@ -100,7 +100,7 @@ define([
       }
 
       // clear placeholder
-      var ph = $("#" + this.htmlObject).empty();
+      var ph = !!this.cleanElementBeforeRender ? $("#" + this.htmlObject).empty() : $("#" + this.htmlObject);
       var me = this;
 
       // Set up defaults for height and width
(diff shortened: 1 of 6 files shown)


core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -100,7 +100,7 @@ define([
       }
 
       // clear placeholder
-      var ph = $("#" + this.htmlObject).empty();
+      var ph = !!this.cleanElementBeforeRender ? $("#" + this.htmlObject).empty() : $("#" + this.htmlObject);
```
dcleao: No need for the `!!` because we're directly on an "if" context.

joaormf: Thanks for the tip

core-js/src/main/javascript/cdf/components/ccc/BaseCccComponent.js
```diff
@@ -199,7 +199,13 @@ define([
         });
       }
 
-      this.chart.render();
+      if (!!this.cleanElementBeforeRender) {
+        this.chart.render();
+      } else {
+        this.chart.render({
+          bypassAnimation: true
+        });
```
dcleao: Idem about the `!!`.
Why the difference in treatment of renders?
I understand the motivation, but it should only be applied on renders after the first one.
Note that `cleanElementBeforeRender` to `true` should be the new default (but we cannot change it now). Like this, animations would never be performed.
The `animate` option should be specified as `false` from the start in these use cases.

PR #983, opened 2018-03-16
[CDE-944] Filter Component - the buttons that collapse groups don't work properly

ricardosilva88:
@pamval @diogofscmariano @afrjorge


ricardosilva88:
@afrjorge ty for the feedback, I wasn't sure if legacy version also requires changes. I'll update the PR

ricardosilva88:
@afrjorge legacy version added to PR

afrjorge:
Seems that you fixed the CSS classes (`expanded` and `collapsed`) used in Group-template.html but you missed the ones in Root-template.html, Root-overlay.html, Root-skeleton.html and Root-header.html. There is also some logic in Root.js to toggle these CSS classes. Should any of these also be changed?
If not, then for the legacy version it seem there is a change missing in template.js: [link here](https://github.com/webdetails/cdf/blob/master/core-js/src/main/javascript/cdf-legacy/components/filter/js/TreeFilter/templates.js#L35)

ricardosilva88:
This problem doesn't affect any Root elements, just the group scope. That's why the change is localized to "group" only.

afrjorge:
Then it seems this is only missing the legacy template change for "Group-template" in template.js.

ricardosilva88:
@afrjorge done, PR updated

PR #989, opened 2018-03-22
[CLEANUP] - Removed usage of undeclared global variable

pamval:


pamval:
@afrjorge please review

afrjorge:
License header year can be updated, there is also still a reference to the invalid var name `_pageSize` in a comment [here](https://github.com/webdetails/cdf/pull/989/files#diff-286720f08789170b305da3b3b9030ddfR284) that can be changed to "page".

pamval:
@afrjorge updated.

PR #991, opened 2018-03-28
[BACKLOG-22244] Fix the CDF components sometimes not refreshing when …

joaormf:
…the component rendering is longer than the refresh period

@marcoslarsen @plagoa @dcleao @afrjorge please review

### DIFF
--- a/core-js/src/main/javascript/cdf/dashboard/RefreshEngine.js
+++ b/core-js/src/main/javascript/cdf/dashboard/RefreshEngine.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2017 Webdetails, a Hitachi Vantara company. All rights reserved.
+ * Copyright 2002 - 2018 Webdetails, a Hitachi Vantara company. All rights reserved.
  * 
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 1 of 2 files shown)


core-js/src/main/javascript/cdf/dashboard/RefreshEngine.js
```diff
@@ -370,16 +384,22 @@ define([], function() {
...
-          //call update, which calls processComponent
-          refreshComponent(info.component);
+          component = info.component;
+
+          //call update, which calls processComponent, but only if not in the dashboard update queue
+          if(!dashboard.updateQueue || dashboard.updateQueue.indexOf(component) == -1) {
+            //setTimeout(function() {
+              refreshComponent(component);
+            //}, 10);
+            reinsertInQueue(component);
```
dcleao: Shouldn't this be in an `else` branch? refresh or reinsert? In that case, because it is not in the queue, we could simply insert in the queue again?

joaormf: I made some subtle changes to the way that the fireResfresh works, and created a reinsertInQueue, with the objective of making the calls to those methods (fireRefresh, and (clearFromQueue+insertInQueue)) non-destructive.

This way I guarantee that the component is in the refresh queue (he should be there anyway because it was explicitly added when the dashboard instance was created), even while the component lifecycle is progressing through it's phases (and hence, that if to not refresh the component if it is in the dashboard update queue).

This last part somehow bleeds some of the logic that should be part of the dashboard lifecycle itself (don't care if it is asked to update multiple times).... but the true is that the behavior that it is applying doesn't work well with the way the refresh engine was built.

**Notice**: the commented lines will be removed.... they were commited by error ( I shall fix that before)

PR #995, opened 2018-04-03
[BACKLOG-22553] Fix test "The Base Ccc Component # allows a dashboard…

joaormf:
… to execute update" failing

@marcoslarsen @plagoa @davidmsantos90 please review

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/Dashboards.RefreshEngine.js
+++ b/core-js/src/main/javascript/cdf-legacy/Dashboards.RefreshEngine.js
@@ -114,9 +114,9 @@ Dashboards.RefreshEngine = function(){// Manages periodic refresh of components
       //tries to fetch the component refresh rate from the data source definition
       if(component.chartDefinition && component.chartDefinition.dataSource) {
         if(dashboard.dataSources && dashboard.dataSources[component.chartDefinition.dataSource]) {
-          if(!Number.isNaN(dashboard.dataSources[component.chartDefinition.dataSource].componentRefreshPeriod) &&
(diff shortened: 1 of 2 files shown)


core-js/src/main/javascript/cdf-legacy/Dashboards.RefreshEngine.js
```diff
@@ -114,7 +114,7 @@ Dashboards.RefreshEngine = function(){// Manages periodic refresh of components
       //tries to fetch the component refresh rate from the data source definition
       if(component.chartDefinition && component.chartDefinition.dataSource) {
         if(dashboard.dataSources && dashboard.dataSources[component.chartDefinition.dataSource]) {
-          if(!Number.isNaN(dashboard.dataSources[component.chartDefinition.dataSource].componentRefreshPeriod) &&
+          if(!_.isNaN(Number.parseInt(dashboard.dataSources[component.chartDefinition.dataSource].componentRefreshPeriod)) &&
```
dcleao: `Number.parseInt` is not supported by IE 11, it seems.
On the other hand, if you use the long supported global `parseInt`, as documented in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt, you should always specify the radix argument, with a value of `10`.

It's usually safer to just use `Number(value)` or `+value` to parse a string into a floating point number. That's because `parseInt` behaves strangely if you have numeric strings with a `0` prefix. If you really need to obtain an integer (not needed in this case), then, floor it down.

joaormf: Nice tip :) 👍

PR #1001, opened 2018-04-20
[CDF-993] - XSS vulnerability in Check Component

YuryBY:
@pamval,  @afrjorge, could you please take a look?

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/Dashboards.Utils.js
+++ b/core-js/src/main/javascript/cdf-legacy/Dashboards.Utils.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2017 Webdetails, a Hitachi Vantara company. All rights reserved.
+ * Copyright 2002 - 2018 Webdetails, a Hitachi Vantara company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
@@ -26,6 +26,12 @@ Dashboards.escapeHtml = function(input) {
   .replace(/"/g,"&#34;");
 };
 
+Dashboards.sanitizeHtml = function(html) {
+  // here is iframe explicitly replaced by script to further sanitizing since sanitizer itself doesn't sanitize iframe tag
+  html = html.replace(/<iframe\b[^>]*>/gi, "<script>").replace(/<\/iframe>/gi, "</script>");
+  html = Sanitizer.sanitize(html);
+  return html;
+};
 
 Dashboards.getPathParameter = function ( url ) {
 
--- a/core-js/src/main/javascript/cdf/components/ToggleButtonBaseComponent.js
+++ b/core-js/src/main/javascript/cdf/components/ToggleButtonBaseComponent.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2017 Webdetails, a Hitachi Vantara company. All rights reserved.
+ * Copyright 2002 - 2018 Webdetails, a Hitachi Vantara company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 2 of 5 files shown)


afrjorge:
Also, unit tests are failing, legacy version also needs to be changed and the license header year updated.

YuryBY:
@afrjorge, @dcleao, the PR is updated. 
Regarding tests failing. I got 
"Failed to run task: 'karma start E:\_w\webdetails\cdf\core-js/src/test/config/javascript/karma.ci.conf.js' failed. (error code 1) " on both ubuntu and windows while running "mvn clean install" so that I need either clean steps to run tests or possibility to run jenkins job on my branch(PR).

afrjorge:
Can you try running `mvn clean install` in the root folder? (`cdf`) There seems to be issues running builds on subfloders in all OSs, and running karma tests on Windows machines.

core-js/src/main/javascript/cdf/components/ToggleButtonBaseComponent.js
```diff
@@ -90,7 +94,7 @@ define(['../lib/jquery', './InputBaseComponent'], function($, InputBaseComponent
         input.appendTo(li);
         li.append($("<label/>")
           .attr({'for': myself.name + i})
-          .html(myArray[i][1]));
+          .Utils.escapeHtml(myArray[i][1]));
```
dcleao: This doesn't look right... Please test the code you propose.

dcleao:
@YuriBy The new approach you took is equivalent to using `text` instead of `html` — all html tags are escaped, and the same loss of functionality is incurred.
 
The CDF Filter component also has the requirement of performing sanitization of the received HTML, so that only a safe subset is allowed.
See the module `cdf/components/filter/HtmlUtils`. It uses the `cdf/lib/sanitizer` library to do the heavy lifting. The former module only adds an additional exception to also escape the `iframe` tag.

I ask you to add a function `sanitizeHtml`, identical to that in `cdf/components/filter/HtmlUtils`, directly to the already existing `cdf/dashboard/Utils` module — the one that contains the `escapeHtml` method you were now using.
Then, use `sanitizeHtml` that one instead of `escapeHtml`.

afrjorge:
You need to add the Sanitizer lib, and its dependencies, to the resources file so it gets loaded in the legacy version. E.g. https://github.com/afrjorge/cdf/commit/b8e494724b30485a9d1e8903bd544bdc1387ad12
Link to legacy sample [here](http://localhost:8080/pentaho/api/repos/%3Apublic%3Aplugin-samples%3Apentaho-cdf%3A30-documentation%3A30-component_reference%3A10-core%3A55-CheckComponent%3Acheck_component.xcdf/generatedContent?debug=true).

YuryBY:
@afrjorge, PR is updated. Provided link is opened and sanitizer is found.

core-js/src/main/javascript/cdf-legacy/Dashboards.Utils.js
```diff
@@ -26,6 +26,12 @@ Dashboards.escapeHtml = function(input) {
   .replace(/"/g,"&#34;");
 };
 
+Dashboards.sanitizeHtml = function(html) {
+  // here is iframe explicitly replaced by script to further sanitizing since sanitizer itself doesn't sanitize iframe tag
+  html = html.replace(/<iframe\b[^>]*>/gi, "<script>").replace(/<\/iframe>/gi, "</script>");
```
mbatchelor: Still would allow <IfRaMe>, <[%00]iframe , <i[%00]fRame . Even this might make it's way through -> %3eiframe . Not sure that this is as effective as you might hope.

YuryBY:
@mbatchelor, currently we have sanitizers with the same code in several locations. Maybe it makes sense to move all of them to one place and to add functionality to cover all scenarios? - or you point is that this should be done in scope of this jira ticket?

core-js/src/main/javascript/cdf-legacy/Dashboards.Utils.js
dcleao: `<IfRaMe>` will not pass as the regular expression is run with the `gi` flags — `i` stands for insensitive. As for the other combinations that you mention, how are they dangerous? Does a browser handle `<i[%00]fRame` as if it were an `iframe`?

PR #1005, opened 2018-04-24
[SP-4232][CDE-943]-Table component that has click action, whenever gets an empty resultset throws an error in jquery.dataTables file.

denisprotopopov:
- fixed
  - apply fix to legacy version
  - updated copyright year


denisprotopopov:
@mbatchelor @mchen-len-son, @pentaho-lmartins, could you please take a look?

PR #1020, opened 2018-07-12
[PPP-4136] Lack of validation on org.pentaho.cdf.CdfApi.buildCdfEmbed…

pentaho-lmartins:
…ContextSecure

- Added OWASP dependency
- Refactored `CdfAPI.buildFullServerUrl` so `protocol` parameter is not written back to the client
- Encoded `serverName` parameter to prevent XSS attach vector
- Cleanup of `CdfApiTest`
- Added more tests to `testBuildFullServerUrl`
- Removed exclusion of transitive dependencies from `cpf-core` and `cpf-pentaho` to reflect guidelines received from @graimundo since these are excluded in the assembly POM

Question: Should we include the `org.owasp.encoder/encoder` dependency with scope `runtime` in the assembly POM so it gets included in the assembly or do we set the scope as `provided` since this is alredy included in the pentaho-platform WAR file.

@dcleao and @graimundo, please review and approve if OK.


pentaho-lmartins:
@dcleao and @graimundo, updated PR with changes to `buildFullServerUrl`. After running some tests and looking at the history of the code, I realized that the parameter `protocol` was being used with the goal to determine the scheme. Protocol version and scheme are two very different things and `protocol` will have the value **HTTP/1.1** even when the scheme is **https**.

pentaho-lmartins:
@graimundo PR updated, removed assembly dependency.

PR #1036, opened 2018-09-18
[BACKLOG-25639] Real time sample dashboard has memory leaks

davidmsantos90:
@webdetails/millenniumfalcon please review.

After doing tests to this CDE sample, they showed that CDF TableComponent was the major culprit in the memory leak. In each component update, when new data was received, the previous DataTable was not disposed, causing the memory leak in the sample.

The chart component also seemed to have some memory leaks, but when talking with @dcleao, we concluded that the extra memory allocation might be caused by protovis internal cache (for text values and such)

### DIFF
--- a/pentaho-js/src/main/javascript/cdf/components/TableComponent.js
+++ b/pentaho-js/src/main/javascript/cdf/components/TableComponent.js
@@ -568,31 +568,28 @@ define([
       var chartDefinition = this.chartDefinition;
       var tablePlaceholder = this.ph;
 
+      if (this.dataTable != null) {
+        this.__removePreviousDataTable();
+      }
 
       tablePlaceholder.trigger('cdfTableComponentProcessResponse');
 
       // Set defaults for headers / types
-      if(typeof chartDefinition.colHeaders === "undefined" || chartDefinition.colHeaders.length === 0) {
-          chartDefinition.colHeaders = json.metadata.map(function(meta) {
+      var hasColumnHeaders = chartDefinition.colHeaders != null && chartDefinition.colHeaders.length > 0;
+      if(!hasColumnHeaders) {
+        chartDefinition.colHeaders = json.metadata.map(function mapColHeaders(meta) {
           return meta.colName;
         });
       }
 
-      if(typeof chartDefinition.colTypes === "undefined" || chartDefinition.colTypes.length === 0) {
-        chartDefinition.colTypes = json.metadata.map(function(meta) {
+      var hasColumnTypes = chartDefinition.colTypes != null && chartDefinition.colTypes.length > 0;
+      if(!hasColumnTypes) {
+        chartDefinition.colTypes = json.metadata.map(function mapColTypes(meta) {
           return meta.colType.toLowerCase();
         });
       }
 
-      var dtData0 = TableComponent.getDataTableOptions(chartDefinition);
-
-      var extraOptions = {};
-      // Build a default config from the standard options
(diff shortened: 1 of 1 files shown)


pentaho-js/src/main/javascript/cdf/components/TableComponent.js
```diff
@@ -492,200 +522,246 @@ define([
...
 
     processTableComponentResponse: function(json) {
-      var myself = this,
-          cd = myself.chartDefinition,
-          extraOptions = {};
+      var myself = this;
+      var chartDefinition = this.chartDefinition;
+      var tablePlaceholder = this.ph;
+
+      if (this.dataTable != null) {
```
dcleao: To preserve backward compatibility the most, probably this should be done after the `trigger` below...

pentaho-js/src/main/javascript/cdf/components/TableComponent.js
```diff
@@ -492,200 +522,246 @@ define([
...
+        this.__removePreviousDataTable();
+      }
 
-      myself.ph.trigger('cdfTableComponentProcessResponse');
+      tablePlaceholder.trigger('cdfTableComponentProcessResponse');
 
       // Set defaults for headers / types
-      if(typeof cd.colHeaders === "undefined" || cd.colHeaders.length == 0) {
-        cd.colHeaders = json.metadata.map(function(i) { return i.colName; });
+      var hasColumnHeaders = chartDefinition.colHeaders != null && chartDefinition.colHeaders.length > 0;
```
dcleao: Backward compatibility: `...colHeaders !== undefined`

pentaho-js/src/main/javascript/cdf/components/TableComponent.js
```diff
@@ -492,200 +522,246 @@ define([
...
+      if(!hasColumnHeaders) {
+        chartDefinition.colHeaders = json.metadata.map(function mapColHeaders(meta) {
+          return meta.colName;
+        });
       }
 
-      if(typeof cd.colTypes === "undefined" || cd.colTypes.length == 0) {
-        cd.colTypes = json.metadata.map(function(i) { return i.colType.toLowerCase(); });
+      var hasColumnTypes = chartDefinition.colTypes != null && chartDefinition.colTypes.length > 0;
+      if(!hasColumnTypes) {
```
dcleao: Idem.

davidmsantos90:
@dcleao update the pull request. please review

pentaho-js/src/main/javascript/cdf/components/TableComponent.js
```diff
@@ -679,16 +680,16 @@ define([
...
+      var dataTableOptions = TableComponent.getDataTableOptions(chartDefinition);
 
       // Build a default config from the standard options
       var extraOptions = {};
       $.each(this.extraOptions != null ? this.extraOptions : {}, function(index, option) {
         extraOptions[option[0]] = option[1];
       });
 
-      return $.extend(chartDefinition.dataTableOptions, dtData0, extraOptions);
+      return $.extend({}, chartDefinition.dataTableOptions, dataTableOptions, extraOptions);
```
dcleao: I also pondered on complaining about this mutating `chartDefinition.dataTableOptions`, but again, to preserve backward compatibility, I think we should keep it the way it was. Nice catch though!

davidmsantos90:
@dcleao you are right, it might break something, reverted the "fix" :+1:

PR #1037, opened 2018-09-19
[BACKLOG-25403] Fix unit test failing

joaormf:


joaormf:
@tgf @marcoslarsen @JoaoRPF and @plagoa please review

PR #1045, opened 2018-10-29
[BACKLOG-26349] Handle close message created from error conditions

JoaoRMF:


JoaoRMF:
Merge with https://github.com/webdetails/cda/pull/292

JoaoRMF:
@marcoslarsen @tgf @plagoa @JoaoRPF please review

core-js/src/main/javascript/cdf/queries/CdaQuery.js
```diff
@@ -399,8 +399,20 @@ define([
 
           myself.getErrorHandler(function (resp, txtStatus, error) {
             errorHandler(resp, txtStatus, error, isDataPush);
-          })(event.data);
+          })(new String(event.data));
```
dcleao: Why would you use a `String` object? If you just want to convert to string, use the `String` function without `new`:
`})(String(event.data));`

JoaoRMF: UnmanagedComponent::failExec it's being checked if the argument has the property "responseText" which doesn't work (ie. exception is thrown) if the argument is a native string, it must .

https://github.com/webdetails/cdf/blob/893ad596bd53b261b1841f741b103370e9d296fa/core-js/src/main/javascript/cdf/components/UnmanagedComponent.js#L402

I think that makes more sense to conform to that "restriction" on the changed code than to change the UnmanagedComponent which could have other implications (the true branch in that if tries to parse the server error).

dcleao: The `in` operator requires an object, and so passing a primitive string fails.
The `failExec` expects an `object`as the value of the `arg` argument. It is documented as:

> An error object or the arguments of a {@link http://api.jquery.com/jquery.ajax/|jQuery.ajax} error callback.

So, definitely, we should be giving it an `Error` object. Otherwise, we could add support for string values to be specified.

So, I'd either change the `failExec` method to support strings, or would change the two call-sites to call it with `new Error(event.data)`.

dcleao: In the latter statement, I'm assuming that `event.data` is always a string, but maybe it is not....

JoaoRMF: That's not a bad solution either... I made it that way (added support for primitive strings in the failExec method, and just pass the event.data (and the new code event.reason usage for the onclose handler) like before.

PR #1047, opened 2018-10-30
[BACKLOG-26349] Handle close message created from error conditions

JoaoRMF:


JoaoRMF:
@pamval please review
@marcoslarsen FYI

PR #1048, opened 2018-11-07
[BACKLOG-26441] Prevent errors after closing the preview window in CD…

JoaoRMF:
…E caused by JSON and _ being discarded when the callback is executed.

@marcoslarsen, @plagoa, @JoaoRPF, @tgf, @dcleao please review

### DIFF
--- a/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
+++ b/core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
@@ -135,9 +135,23 @@
           var isDataPush = !isFirstTime;
           isFirstTime = false;
 
-          myself.getSuccessHandler(function (data) {
-            return successHandler(data, isDataPush);
-          })(JSON.parse(event.data));
+          // If instruction introduced to safeguard against IE 11 and Edge 24 not closing
+          // the websocket after the CDE preview (iframe) is closed, resulting
+          // in a strange window state where some should-exist objects are undefined
+          if(JSON) {
+            myself.getSuccessHandler(function (data) {
+              return successHandler(data, isDataPush);
+            })(JSON.parse(event.data));
+          } else {
+            // when in the strange situation where the JSON object is not defined/null
+            // we give our best shot in trying to close the websocket
+            try {
+              //1001 - Going Away
+              myself.websocket.close(1001);
+            } catch(err) {
+              //empty on purpose
+            }
+          }
         };
 
         this.websocket.onerror = function (event) {
--- a/core-js/src/main/javascript/cdf/queries/CdaQuery.js
+++ b/core-js/src/main/javascript/cdf/queries/CdaQuery.js
@@ -388,9 +388,23 @@ define([
           var isDataPush = !isFirstTime;
           isFirstTime = false;
 
-          myself.getSuccessHandler(function (data) {
-            return successHandler(data, isDataPush);
-          })(JSON.parse(event.data));
+          // If instruction introduced to safeguard against IE 11 and Edge 24 not closing
+          // the websocket after the CDE preview (iframe) is closed, resulting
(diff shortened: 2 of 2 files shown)


JoaoRMF:
Dev note...

I check for JSON and underscore existence because if only JSON was tested the exception would be raised downstream because of the underscore.

core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
```diff
@@ -135,9 +135,13 @@
           var isDataPush = !isFirstTime;
           isFirstTime = false;
 
-          myself.getSuccessHandler(function (data) {
-            return successHandler(data, isDataPush);
-          })(JSON.parse(event.data));
+          // we need to check if the JSON and underscore exist because
+          // after closing the preview on CDE they cease to exist
+          if(JSON && _) {
```
dcleao: Do you mean, the symbols remain defined globally but have a null or undefined value?
Isn't there some other way to detect that the "window" has been disposed of?

JoaoRMF: I would say that the JSON and underscore are defined in a specific closure created for the preview. Since the preview is closed, the closure is destroyed, but since the websocket messages are async the onmessage handler has a backlog of messages to process, which on Edge seem to be kept on memory even after the closure is destroyed.

I would say that checking if the dashboard was disposed or something like this would be "prettier" but since the entire closure is destroyed I'm not sure that we can gain access to the dashboard.

Maybe we could store an object which is copied into the handler function, but then this could create a memory leak.

JoaoRMF:
Yes, I would say that this is only for master.

pamval:
No, this issue is marked as an 8.2 issue. If this is _not_ an issue for 8.2 because it only happens in an unsupported Edge version, please let us know so that we remove it from the 8.2 board

core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
dcleao: JSON is a native object, provided by the browser itself.
Underscore is provided by:
* legacy dashboard: the underscore script (or cdf bundle), and is published in the global scope at script load
* amd dashboard: the `CdaQuery`'s AMD module's closure

So, there must be something wrong for these to become _nully_ valued (and still defined) — either a browser bug or something else which we're missing.

If this is a bug of a specific browser, then I think that that should be noted beside the code.

JoaoRMF:
@pamval, the issue only happens in 8.2 but on a newer Edge version which is not supported.

core-js/src/main/javascript/cdf-legacy/queries/coreQueries.js
JoaoRMF: It should be like a normal rendering, but the difference is that when the iframe is unloaded it seams that the websocket is not being disposed and is kept open. As per the specification (https://www.w3.org/TR/websockets/#garbage-collection) when the document is disposed, then the websocket close handshaking should start, and it doesn't.

I have assembled a simple html webpage with an iframe, and the websocket onmessage handler is still being triggered after the body onunload event is executed (this in Edge and IE 11).

I also noticed that the console object was not available after the body unload, inner functions used in my javascript block weren't available... but the handler code was still being called.

dcleao: I'm fine with the solution you've found.
Please, just add a comment justifying the odd construction, and, perhaps, pointing to the related issue :-).

JoaoRMF: The issue was raised here: https://jira.pentaho.com/browse/BACKLOG-26441
I will add information on the issue itself, regarding the conclusions that we got to, explain the code, and upload the demo page that I assembled.

PR #1059, opened 2019-01-08
[CDF-1024] CPF InterPluginCall changed causing CDF autoIncludes to fail.

usbrandon:
The second parameter of an InterPluginCall expects a valid method that exists in the class.
Originally someone passed in "" or no method, this caused the method compare to fail and gave a false report that CDA was not installed.
listQueries is a valid method for CDA, and thus will have the check work and return true.

AutoInclude logic works and was unmodified.


mbatchelor:
@usbrandon - Hey Brandon, thanks for the contribution!!! You rock.

PR #1064, opened 2019-03-13
[BACKLOG-27092] - On IE11 and Edge the setTimeout requires to be boun…

marcoslarsen:
…ded to window in the blockUi execution context to prevent "Invalid calling object" exception.


marcoslarsen:
@tgf @JoaoRMF @JoaoRPF @plagoa Please review.

PR #1071, opened 2019-04-04
[BACKLOG-28478] Fix incorrect call to Analyzer on ExecuteAnalyzerReportComponent

davidmsantos90:
@webdetails/millenniumfalcon please review

 - Mostly cleanup work, the issue was the component not passing the correct path options (solution, path and action) to the api call.

### DIFF
--- a/pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
+++ b/pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2017 Webdetails, a Hitachi Vantara company. All rights reserved.
+ * Copyright 2002 - 2019 Webdetails, a Hitachi Vantara company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
  * of the Mozilla Public License, Version 2.0, or any later version. You may not use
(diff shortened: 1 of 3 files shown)


pentaho-js/src/main/javascript/cdf-legacy/components/Pentaho.Analyzer.js
```diff
@@ -109,29 +109,24 @@ var ExecuteAnalyzerComponent = AnalyzerComponent.extend({
...
-        }
-        if ( parameters.action ) {
-            $.extend( path, {action: parameters.action});
-        }
-        delete parameters.solution;
-        delete parameters.path;
-        delete parameters.action;
-        $.extend( parameters, {ts: new Date().getTime()});
+
+        parameters.ts = new Date().getTime();
```
dcleao: Are you sure that the following removed lines are not required?
```js
       delete parameters.solution;
        delete parameters.path;
        delete parameters.action;
```

davidmsantos90: Yes, they were supposed to be removing this values from the object, to avoid passing them as query params to the API request.

But this values were populated using the component's (solution, path and action) properties in the `getOptions` method. (In other components, because in ExecuteAnalyzerComponent we were ignoring their values).

Now they are never there, so we don't need to delete those values

PR #1084, opened 2019-07-10
[CDE-1005] CDE dashboard can't be localized at country level

LeonardoCoelho71950:
Our environment locale is being lowercased here: https://github.com/pentaho/pentaho-platform-plugin-common-ui/blob/master/impl/client/src/main/javascript/web/pentaho/environment/impl/Environment.js#L64 

So, we need to do the same here, if we want a successful match. But we need to return the supported cased locale to download the wanted properties file.

### DIFF
--- a/core-js/src/main/javascript/lib/i18n/cdf.jquery.i18n.js
+++ b/core-js/src/main/javascript/lib/i18n/cdf.jquery.i18n.js
@@ -1,5 +1,5 @@
 /*!
- * Copyright 2002 - 2017 Webdetails, a Hitachi Vantara company. All rights reserved.
+ * Copyright 2002 - 2019 Webdetails, a Hitachi Vantara company. All rights reserved.
  *
  * This software was developed by Webdetails and is provided under the terms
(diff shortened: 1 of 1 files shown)


core-js/src/main/javascript/lib/i18n/cdf.jquery.i18n.js
```diff
@@ -102,12 +102,12 @@ define([
         var supportedLocale = lines[i];
 
         var lang = supportedLocale.substr(0, supportedLocale.indexOf("="));
-        if (lang === locale && result == null) {
-          result = locale;
+        if (lang.toLowerCase() === locale && result == null) {
```
dcleao: I think it would be best to create a `langLower` variable.

LeonardoCoelho71950: Thanks for the feedback

LeonardoCoelho71950: I made your suggestions for this.
I am currently analyzing the common-ui file.

PR #1100, opened 2019-10-10
[PDB-2060] - Xactions cannot be used in Dashboard Designer

bcostahitachivantara:
@pentaho-lmartins 
@ssamora

### DIFF
--- a/core-js/src/main/javascript/cdf/dashboard/Utils.js
+++ b/core-js/src/main/javascript/cdf/dashboard/Utils.js
@@ -734,6 +734,52 @@ define([
     return "rgb(" + rgb.join(",") + ")";
   };
 
+  /**
+   * Determines if `path` ends with `action`,
+   * immediately preceded by a `\` or `/` separator character.
+   *
+   * Pattern to test: `<path> === <basepath>[\/]<action>`.
+   *
+   * This implementation, in particular, strives to not allocate memory.
+   *
+   * @param {*} path - The path.
+   * @param {*} action - The action.
+   * @return {boolean} `true` if yes; `false`, otherwise.
+   */
+  Utils.pathIncludesAction = function (path, action) {
+
+    if(typeof path !== "string" || typeof action !== "string") {
+      return false;
+    }
+
+    if(action.length >= path.length) {
+      return false;
+    }
+
+    if( "\\/".indexOf(path.substr(-1, 1)) >= 0) {
+      return false;
+    }
+
+    if(!Utils.endsWith(path, action)) {
+      return false;
+    }
+
+    var separator = path.substr(-(action.length + 1), 1);
+
+    return "\\/".indexOf(separator) >= 0;
+  };
+
+  /**
+   * Determines if one string ends with another.
+   *
+   * @param {string} test - The string being tested to end with `suffix`.
+   * @param {string} suffix - The suffix string.
+   * @return {boolean} `true` if `test` ends with `suffix`; `false`, otherwise.
+   */
+  Utils.endsWith = function(test, suffix) {
+    return test === suffix || test.slice(-suffix.length) === suffix;
+  };
+
   return Utils;
 
 });
--- a/core-js/src/main/javascript/cdf/dashboard/Utils.js
+++ b/core-js/src/main/javascript/cdf/dashboard/Utils.js
@@ -752,6 +752,8 @@ define([
       return false;
     }
 
+    // Validate if path ends with '/' or '\'
+    // Example: Utils.pathIncludesAction( "/a.ext/", "a.ext/" ) should be false
     if( "\\/".indexOf(path.substr(-1, 1)) >= 0) {
       return false;
     }
--- a/core-js/src/main/javascript/cdf/dashboard/Utils.js
+++ b/core-js/src/main/javascript/cdf/dashboard/Utils.js
@@ -752,10 +752,6 @@ define([
       return false;
     }
 
-    if(action.length >= path.length) {
-      return false;
-    }
-
     if( "\\/".indexOf(path.substr(-1, 1)) >= 0) {
       return false;
     }
--- a/core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
+++ b/core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
@@ -116,5 +116,27 @@ define(["cdf/dashboard/Utils", "cdf/lib/CCC/cdo"], function(Utils, cdo) {
       expect(Utils.ev("1")).toBe("1");
       expect(Utils.ev(function() { return "1"; })).toBe("1");
     });
+
+    /**
+     * ## The Utils class # verify if pathIncludesAction evaluates well
+     */
+    it( "evaluate if path includes action", function () {
+      expect( Utils.pathIncludesAction( "a.ext", "a.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext", "a.ext" ) ).toBe( true );
+      expect( Utils.pathIncludesAction( "/aa.ext", "a.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/aa.ext", null ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( "/a/a.ext/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( null, "a.ext") ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", undefined ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/short/path/a.ext", "very_long_long_long_long_long_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/very/long/long/long/long/long/long/path/a.ext", "short_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext/", "a.ext/" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext\\", "a.ext\\" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "/a.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a\\a.ext", "a\\a.ext" ) ).toBe( true );
+      expect( Utils.pathIncludesAction( "\\a.ext", "a.ext" ) ).toBe( true );
+    });
   });
 });
--- a/pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
+++ b/pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
@@ -14,8 +14,9 @@
 define([
   './XactionComponent.ext',
   '../lib/jquery',
-  './BaseComponent'
-], function(XactionComponentExt, $, BaseComponent) {
+  './BaseComponent',
+  '../dashboard/Utils'
+], function(XactionComponentExt, $, BaseComponent, Utils) {
 
   return BaseComponent.extend({
     update: function() {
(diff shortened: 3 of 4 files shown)


buildguy:
### :gear: Build finished in 2m 41s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|71|0|0|

|[Details](http://wingman2.pentaho.net/job/projects/job/cdf-plugin/job/PR-1100/1/testReport)|
|----|


:information_source:  _This is an automatic message_

pentaho-lmartins:
@dcleao, can you please review this one?

pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
```diff
@@ -49,7 +49,20 @@ define([
...
                                   " width=\"100%\" />";
           var iframe = $(xactionIFrameHTML);
-          var url = XactionComponentExt.getCdfXaction(this.path, this.action, this.solution) + "&wrapper=false";
+          var actionIncluded = function(path, action) {
+            //check if path ends with action prefixed with '\' or '/'
+            return (typeof path == "string") && (typeof action == "string")
+                && (path.length > action.length)
+                && (path.lastIndexOf(action) == (path.length - action.length))
+                && ("\\/".indexOf(path.substr(-action.length-1, 1))>=0);
+          };
```
dcleao: This is great. Very compact :-) Still, I'd prefer a more readable version of this code, which relied on common operations such as `endsWith`. I know this `String` method cannot yet be safely used, but we can roll our own.

You could place the following two functions at the end of the file, just before the closing `});` of the AMD module:

```js
/**
 * Determines if `path` ends with `action`, 
 * immediately preceded by a `\` or `/` separator character.
 * 
 * Pattern to test: `<path> === <basepath>[\/]<action>`.
 * 
 * This implementation, in particular, strives to not allocate memory.
 * 
 * @param {*} path - The path.
 * @param {*} action - The action.
 * @return {boolean} `true` if yes; `false`, otherwise.
 */
function pathIncludesAction(path, action) {

  if(typeof path !== "string" || typeof action !== "string") {
    return false;
  }

  if(!endsWith(path, action)) {
    return false;
  }

  var separator = path.substr(-(action.length + 1), 1);
  
  return "\\/".indexOf(separator) >= 0;
}

/**
 * Determines if one string ends with another.
 * 
 * @param {string} test - The string being tested to end with `suffix`.
 * @param {string} suffix - The suffix string.
 * @return {boolean} `true` if `test` ends with `suffix`; `false`, otherwise.
 */
function endsWith(test, suffix) {
  return test === suffix || test.slice(-suffix.length) === suffix;
}
```

Documentation is also a good thing to do, even if sometimes it doesn't seem to add much.
It states our intent and helps to validate our assumptions. It allows future folks to know what was on our mind.

Also, note the proposed name, `pathIncludesAction`, instead of `actionIncluded`. If you really prefer something more like you original formulation, then, maybe something like `isActionIncluded` or `doesPathIncludeAction` would still be better.

pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
```diff
@@ -49,7 +49,20 @@ define([
...
+          var url;
+          if (actionIncluded(myself.path, myself.action)) {
+            url = XactionComponentExt.getCdfXaction(myself.path, "", myself.solution) + "&wrapper=false";
+          } else {
+            url = XactionComponentExt.getCdfXaction(myself.path, myself.action, myself.solution) + "&wrapper=false";
+          }
+
```
dcleao: Here, I'd prefer to not repeat the longer expression, which calculates the final URL.
It could be like:
```js
var actionEf = pathIncludesAction(this.path, this.action) ? "" : this.action;

var url = XactionComponentExt.getCdfXaction(this.path, actionEf, this.solution) +
     "&wrapper=false";
```

Additionally, note that it is not necessary to use the `myself` variable, instead of `this`, as this code lies _directly_ in the body of the `update` method.

pentaho-js/src/test/javascript/cdf/components/XactionComponent-spec.js
```diff
@@ -56,5 +60,22 @@ define([
 
       dashboard.update(xactionComponent);
     });
+
+    it("Check if update calls getCdfXaction method with correct parameters", function(done) {
```
dcleao: First, ideally, the title of a test case _should_ start with "should".
Something like "should call getCdfXAction method with correct parameters".

Then, the test title _should_ describe what is effectively being tested. In this case, you are testing one of the cases, where "`path` ends with `action` and `action` is preceded by `/`".

So, a test title like the following would be more appropriate: "should call `getCdfXAction` with an empty action, when `path` ends with `action` preceded by `/`".

Lastly, you are not testing other main case, in which `path` does _not_ end with `action`. I think you should exercise the following (some, corner'y) cases:
- `action` is `null`
- `action` is `undefined`
- `action` is ""
- `path` equals `action`,
- `path` is shorter than `action`
- `path` is longer but does not end with `action`
- `path` ends with `action` and `action` is preceded by `/`
- `path` ends with `action` and `action` is preceded by `\`
- `path` ends with `action` but `action` is preceded by something other than `/` or `\`

To be able to test these different cases, you should refactor the tests to not create the dashboard directly within the body of the `describe` block (which is always bad practice anyway, unless you're declaring a constant...).

You can define a function which creates a dashboard, given a path argument. Then, call it within the individual test cases (i.e. `it` functions).

buildguy:
### :gear: Build finished in 2m 39s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|71|0|0|

|[Details](http://wingman2.pentaho.net/job/projects/job/cdf-plugin/job/PR-1100/2/testReport)|
|----|


:information_source:  _This is an automatic message_

core-js/src/main/javascript/cdf/dashboard/Utils.js
```diff
@@ -734,6 +734,52 @@ define([
...
+   * @param {*} action - The action.
+   * @return {boolean} `true` if yes; `false`, otherwise.
+   */
+  Utils.pathIncludesAction = function (path, action) {
+
+    if(typeof path !== "string" || typeof action !== "string") {
+      return false;
+    }
+
+    if(action.length >= path.length) {
```
dcleao: Why are you excluding the case where they are equal? `action` is actually included in `path` in that case — a legitimate use case imo.

Otherwise, this is an optimization which you can be simply delegated to `endsWith`.

core-js/src/main/javascript/cdf/dashboard/Utils.js
```diff
@@ -734,6 +734,52 @@ define([
...
+      return false;
+    }
+
+    if( "\\/".indexOf(path.substr(-1, 1)) >= 0) {
```
dcleao: This is a special case that would otherwise be immediately detected by `endsWith` (assuming `action` does not itself end with a `/` or `\`). So why repeat it?

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
```diff
@@ -116,5 +116,27 @@ define(["cdf/dashboard/Utils", "cdf/lib/CCC/cdo"], function(Utils, cdo) {
...
+      expect( Utils.pathIncludesAction( "/aa.ext", "a.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/aa.ext", null ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( "/a/a.ext/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( null, "a.ext") ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", undefined ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/short/path/a.ext", "very_long_long_long_long_long_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/very/long/long/long/long/long/long/path/a.ext", "short_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext/", "a.ext/" ) ).toBe( false );
```
dcleao: Again, disagree with this one. I don't think you need to know the structure of action.

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
```diff
@@ -116,5 +116,27 @@ define(["cdf/dashboard/Utils", "cdf/lib/CCC/cdo"], function(Utils, cdo) {
       expect(Utils.ev("1")).toBe("1");
       expect(Utils.ev(function() { return "1"; })).toBe("1");
     });
+
+    /**
+     * ## The Utils class # verify if pathIncludesAction evaluates well
+     */
+    it( "evaluate if path includes action", function () {
+      expect( Utils.pathIncludesAction( "a.ext", "a.ext" ) ).toBe( false );
```
dcleao: I disagree with this one. Already explained above.

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
```diff
@@ -116,5 +116,27 @@ define(["cdf/dashboard/Utils", "cdf/lib/CCC/cdo"], function(Utils, cdo) {
...
+      expect( Utils.pathIncludesAction( "/aa.ext", null ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( "/a/a.ext/a.ext", "a.ext") ).toBe( true );
+      expect( Utils.pathIncludesAction( null, "a.ext") ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", undefined ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a/a.ext", "" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/short/path/a.ext", "very_long_long_long_long_long_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/very/long/long/long/long/long/long/path/a.ext", "short_action.ext" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext/", "a.ext/" ) ).toBe( false );
+      expect( Utils.pathIncludesAction( "/a.ext\\", "a.ext\\" ) ).toBe( false );
```
dcleao: Idem.

pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
```diff
@@ -49,7 +50,10 @@ define([
                                   " height=\"100%\"" +
                                   " width=\"100%\" />";
           var iframe = $(xactionIFrameHTML);
-          var url = XactionComponentExt.getCdfXaction(this.path, this.action, this.solution) + "&wrapper=false";
+          var url = XactionComponentExt.getCdfXaction(myself.path,
+              Utils.pathIncludesAction(this.path, this.action) ? "" : this.action || "",
```
dcleao: I don't think we should need to know the precedence order to such detail, by heart. Unusual cases should just use parenteses, imo.

Please, change the last bit to `(this.action || "")` (assuming I guessed the correct precedence order).

Also, if a variable was defined to hold the effective `action` value, it would be much easier to read.

core-js/src/main/javascript/cdf/dashboard/Utils.js
bcostahitachivantara: It was a mistake, i was thinking only in case where path includes action.

pentaho-js/src/main/javascript/cdf/components/XactionComponent.js
bcostahitachivantara: I agree with your suggestion about parentheses ;)

core-js/src/main/javascript/cdf/dashboard/Utils.js
bcostahitachivantara: This condition is exactly to prevent actions that finish with `/` or `\`. 

I found this when i was running these tests:
```
expect( Utils.pathIncludesAction( "/a.ext/", "a.ext/" ) ).toBe( false );
expect( Utils.pathIncludesAction( "/a.ext\\", "a.ext\\" ) ).toBe( false );
```

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
bcostahitachivantara: I didn't understand what you mean

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
bcostahitachivantara: Same above

core-js/src/test/javascript/cdf/dashboard/Utils-spec.js
bcostahitachivantara: Same above

buildguy:
### :gear: Build finished in 2m 39s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|71|0|0|

|[Details](http://wingman2.pentaho.net/job/projects/job/cdf-plugin/job/PR-1100/3/testReport)|
|----|


:information_source:  _This is an automatic message_

buildguy:
### :gear: Build finished in 2m 50s with command:
```console
mvn clean verify -B -e -Daether.connector.resumeDownloads=false -amd \
-Daudit -P !assemblies
```
:ok_hand: All tests passed!
 ---

|Total|Failed|Skipped|
|---|---|---|
|71|0|0|

|[Details](http://wingman2.pentaho.net/job/projects/job/cdf-plugin/job/PR-1100/4/testReport)|
|----|


:information_source:  _This is an automatic message_

PR #1171, opened 2023-11-09
[BISERVER-14940] - Reports with Single and Multi Value Cascading Parameters will throw errors on the Pentaho Server - Request with "//" will throw Spring Security Error

andreramos89:
In the class Dashboard.legacy.js, in the function "pentahoServiceAction", when the path and action are both undefined, the request will always fail. Also, It will generate a url with double slash.

@dcleao 

@renato-s 
@smmribeiro 
@bcostahitachivantara


buildguy:
### :white_check_mark: Build finished in 3m 14s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
:ok_hand: All tests passed!
 
Tests run: **71**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/cdf-plugin/job/PR-1171/1/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

core-js/src/main/javascript/cdf/components/BaseComponent.js
```diff
@@ -391,7 +391,7 @@ define([
               arr[val[0]] = val[1];
             });
             jXML = this.dashboard.parseXActionResult(myself, this.dashboard.urlAction(this.url, arr));
-          } else {
+          } else if(this.path!=null && this.solution!=null) {
```
dcleao, line 394: It appears that `solution` is being ignored (is (ultra-)deprecated), so it may be that some are passing a nully value, and thus it should continue working regardless of the value of solution.

The issue currently being solved, resulting in a URL with a `//` segment,  happens only when both path and action are nully.

So, I believe the condition should be `this.path != null || this.action != null`.

core-js/src/main/javascript/cdf/dashboard/Dashboard.legacy.js
```diff
@@ -176,7 +176,6 @@ define([
      */
     pentahoServiceAction: function(serviceMethod, returntype, solution, path, action, params, func) {
       // execute an Action Sequence on the server
-
```
dcleao, line 179: No real changes in this file.

dcleao (changes requested):

core-js/src/main/javascript/cdf/components/BaseComponent.js
andreramos89: I may have misunderstood the point, but I believe it should be "if(this.path && this.action)" as having one or both missing, will result in a failing request

dcleao: We are not sure what current uses exist of these functions. We're only sure that it will fail if both are nully. Hence the proposed condition. Your proposal could cause regressions on cases where only one of path or action is being specified.

andreramos89: That is correct. I was confused

dcleao (approved):

bcostahitachivantara (approved):
Looks good!

bcostahitachivantara:
If you want, you could also write a comment explaining why we do not need to do the request and mention the jira issue

smmribeiro (approved):

core-js/src/main/javascript/cdf-legacy/components/core.js
```diff
@@ -168,7 +168,8 @@ BaseComponent = Base.extend(Backbone.Events).extend({
             arr[val[0]]=val[1];
           });
           jXML = Dashboards.parseXActionResult(myself, Dashboards.urlAction(this.url, arr));
-        } else {
+        } else if(this.path || this.action) {
+          //BISERVER-14940 if path and action are both null, the request will fail, as it does not have enough parameters for the request
```
dcleao, line 172: Also, in this case, it happens because the Prompting API already provides the `valuesArray` filled in, and there is no need to fetch the values from a service. However, in some cases (of dependent selects), the array is empty.

dcleao:
> If you want, you could also write a comment explaining why we do not need to do the request and mention the jira issue

https://github.com/webdetails/cdf/pull/1171/commits/dff256ea0cbe5a15a339b8320e941f1274f03e87#r1389546771

buildguy:
### :white_check_mark: Build finished in 3m 30s
Build command:
```console
mvn clean verify -B -e -Daudit -amd
```
---
:ok_hand: All tests passed!
 
Tests run: **71**, Failures: **0**, Skipped: **0** &nbsp;&nbsp; [Test Results](https://wingman.orl.eng.hitachivantara.com/job/projects/job/cdf-plugin/job/PR-1171/2/testReport)

 ---
 
 

:information_source:  _This is an automatic message_

